PR #31, opened 2017-07-10
chore: link to article about semantic commits

dchest:
https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md


dchest:
(change link to https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md)

PR #39, opened 2017-08-02
Update deps

dchest:
Updates:

* `electron-builder`
* `@peerio/updater`
* `@peerio/desktop-release-builder`

Deletes:

* `electron-builder-squirrel-windows` (we don't use it)

Adds:

Infrastructure for enabling @peerio/updater for Windows , but doesn't enable it yet — let's release with Linux only to better test it.

Tested on my machine, builds work and Windows signing works too.

NOTE: we need at least 2 staging releases to test that nothing was broken in electron-updater (it wasn't updated, but the builder itself updated). This also allow us to test Linux updater.


package.json
```diff
@@ -6,7 +6,7 @@
     "precommit": "npm run validate",
     "postmerge": "npm update",
     "postcommit": "git fetch --prune origin '+refs/tags/*:refs/tags/*'",
-    "postinstall": "install-app-deps",
+    "postinstall": "electron-builder install-app-deps",
```
anri-asaturov: you sure u didn't miss `&&`  `;` or `||` between commands?

dchest: no, it's a subcommand of `electron-builder`:


Before:

```
Warning: Please use as subcommand: electron-builder install-app-deps
```

PR #45, opened 2017-08-11
Feat win32 debug build

seavan:
npm-run-all -l -n compile -p ...

putting any tasks before "-p" switch in win environment causes crash, use "&&" instead.

"mkdir" signature is different on windows, and cannot be easily aliased. use "mkdirp" from npm instead.


package.json
```diff
@@ -8,10 +8,10 @@
     "postcommit": "git fetch --prune origin '+refs/tags/*:refs/tags/*'",
     "postinstall": "electron-builder install-app-deps",
     "commitmsg": "validate-commit-msg",
-    "start": "npm-run-all -l -n compile -p code:watch icebear:watch style:watch devstyle:watch electron:run",
+    "start": "npm run compile && npm-run-all -l -p code:watch icebear:watch style:watch devstyle:watch electron:run",
     "start-latest": "git pull && npm install && npm update && npm start",
-    "compile": "npm-run-all -l -n cleanup -p static:copy code:compile icebear:compile style:compile devstyle:compile react-toolbox",
-    "cleanup": "rm -rf app/build && mkdir -p app/build",
+    "compile": "npm run cleanup && npm-run-all -l -n -p static:copy code:compile icebear:compile style:compile devstyle:compile react-toolbox",
+    "cleanup": "rm -rf app/build && mkdirp app/build",
```
dchest: Does `rm` work for you? I used `rimraf` from npm instead so that it would work on Windows.

PR #46, opened 2017-08-11
Feat sanitize filenames

seavan:
sanitize default filenames


package.json
```diff
@@ -246,5 +247,7 @@
       "changelog": true
     }
   },
-  "dependencies": {}
+  "dependencies": {
+    "sanitize-filename": "1.6.1"
```
anri-asaturov: If this is a prod dependency, it should go to app/package.json

src/helpers/file.js
```diff
@@ -2,10 +2,11 @@ const electron = require('electron').remote;
 const fs = require('fs');
 const _ = require('lodash');
 const path = require('path');
+const sanitize = require('sanitize-filename');
 
 function requestDownloadPath(fileName) {
     return new Promise((resolve, reject) => {
-        let p = fileName;
+        let p = sanitize(fileName, { replacement: '_' });
```
anri-asaturov: @dchest please review

dchest: Looks good, but it would be nice if it also sanitized when uploading so that there's no surprises with filenames.

dchest: Maybe do this in the SDK?

anri-asaturov:
Ok. Merging this, not sure about 'for upload' let's discuss with product team.

PR #51, opened 2017-08-14
feat: file signature error

seavan:


src/ui/chat/components/InlineFiles.jsx
```diff
@@ -37,6 +39,28 @@ class InlineFiles extends React.Component {
         downloadFile(file);
     }
 
+    get signatureErrorDialog() {
+        const hide = () => (this.showSignatureErrorDialog = false);
```
anri-asaturov: This component is performance-critical, move hide function out so it doesn't get re-created every render.

src/ui/chat/components/InlineFiles.jsx
```diff
@@ -46,20 +70,29 @@ class InlineFiles extends React.Component {
...
-                            return (<div className="invalid-file" key={f}>
+                            return (<div className="unknown-file" key={f}>
                                 {t((this.giveUp || gaveUpMap[f]) ? 'error_fileRemoved' : 'title_processing')}
                                 {this.giveUp || gaveUpMap[f]
                                     ? null
                                     : <ProgressBar type="linear" mode="indeterminate" />
                                 }
                             </div>);
                         }
+                        if (true || file.signatureError) {
```
anri-asaturov: if true ?

src/ui/chat/components/InlineFiles.jsx
```diff
@@ -46,20 +70,29 @@ class InlineFiles extends React.Component {
...
+                            return (
+                                <div className="invalid-file" key={f} data-id={f} onClick={() => this.showSignatureError()}>
```
anri-asaturov: ```onClick={() => this.showSignatureError()}```
nope
```onClick={this.showSignatureError}```
and make showSignatureError an arrow function

File: src/ui/chat/components/InlineFiles.jsx [hunk 4]
```diff
@@ -68,6 +101,7 @@ class InlineFiles extends React.Component {
                         </div>);
                     })
                 }
+                {this.signatureErrorDialog}
```
anri-asaturov: So we're rendering the same dialog for every inline file even though it doesn't contain any file-specific data? 
Do one of these instead:
1. Move this line inside `<div className="invalid-file">` so it renders only for invalid files at least.
2. Make dialog global, render it in Chat.jsx or MessageList.jsx and make it trigger via uiState (see contactDialog it's doing the same)

PR #52, opened 2017-08-16
fix: channelLimit === null means no limit

seavan:


anri-asaturov:
I will not merge this because if you look at how file size and ghost limit is done - you'll see why.

PR #58, opened 2017-09-25
feat(chats): bolded chat names when unreads

flohdot:
unreads are now bold


src/ui/chat/components/ChatList.scss
```diff
@@ -33,6 +38,8 @@
       @extend %flex-align-center;
       height: 32px;
     }
+
+
```
anri-asaturov: These empty lines not removed by linter tell me you have hook issues again ;)

flohdot: ooops

PR #59, opened 2017-09-27
feat(chat): DM sidebar simplified, toggle only shows when recent files exist

suprko:


src/ui/chat/components/sidebar/ChatSideBar.jsx
```diff
@@ -40,14 +40,20 @@ class ChatSideBar extends React.Component {
...
-                <div>
-                    <ChatNameEditor showLabel tabIndex="-1" readOnly={!canIAdmin} />
-                    {isChannel ? <ChatNameEditor showLabel tabIndex="-1" purpose readOnly={!canIAdmin} /> : null}
-                </div>
+                {isChannel ?
+                  <div>
+                    <div className="title">{t('title_About')}</div>
+                    <div>
+                        <ChatNameEditor showLabel tabIndex="-1" readOnly={!canIAdmin} />
+                        {isChannel ? <ChatNameEditor showLabel tabIndex="-1" purpose readOnly={!canIAdmin} /> : null}
```
anri-asaturov: There's aready a check for isChannel on the line 47, this check is excessive.

src/ui/chat/components/sidebar/ChatSideBar.jsx
```diff
@@ -40,14 +40,20 @@ class ChatSideBar extends React.Component {
         const chat = chatStore.activeChat;
         if (!chat) return null;
         const { isChannel, canIAdmin, canILeave } = chat;
+        if (!isChannel && !chat.recentFiles.length) return null;
```
anri-asaturov: So, channels don't need sidebars? @suprko

src/ui/chat/components/sidebar/ChatSideBar.jsx
suprko: Ah right, a previous idea that's now redundant. (Keep title but make read only). I'll fix it

PR #61, opened 2017-10-02
Refactor MessageInput component to isolate Quill-specific functionality

lostfictions:
As the first step in swapping Quill out for ProseMirror, I've refactored out the Quill-specific parts of MessageInput into their own child component.

This PR also gets rid of the ComposeMessage superclass MessageInput was inheriting from, since Quill functionality was split across both components but it wasn't really being shared with anything else. (The other ComposeMessage subclass was a Mail-related component that likely wouldn't work anymore anyway since it's been dead code for a while -- I've just made it throw if anyone tries to use it now.)

One potentially weird thing about this PR is that MessageInputQuill uses React 16's new ability to return an array of components instead of a single wrapped component -- that's because I tried to keep as much common functionality (file picker, thumbs-up/ack) in the parent but had to put the emoji picker (and picker button) in the Quill-specific file along with the input box proper. But everything's set up for all that stuff to render in a flat hierarchy with flex, so this seemed less the less disruptive option.


src/ui/chat/components/MessageInput.jsx
```diff
@@ -404,62 +94,33 @@ class MessageInput extends React.Component {
...
                     }
-                    <IconButton icon="mood" disabled={this.emojiPickerVisible} onClick={this.showEmojiPicker} />
+                    <IconButton
+                        disabled={!chat || !chat.canSendAck}
+                        icon="thumb_up"
+                        onClick={this.props.onAck}
+                        className="color-brand"
+                    />
 
-                    {this.text === ''
```
lostfictions: `this.text` was unused and always empty, so we just always render the thumbs-up button now :)

anri-asaturov:
Great job, I don't think returning array is weird, it's a nice new feature reducing noise, we should use it.

PR #64, opened 2017-10-13
Fix new dm

seavan:
cleaning up unused code
fixes for bugs which Eren found
sidebar toggle state moved to ui-state


src/stores/ui-store.js
```diff
@@ -70,6 +73,13 @@ class UIStore {
                 });
             }
         });
+
+        TinyDb.user.getValue(SIDEBAR_STATE_KEY).then(isOpen => {
```
anri-asaturov: Add chatSideBarIsOpen key to UIStore#prefs please, it's gonna take care of save/load and you can access it with `UIStore#prefs.chatSideBarIsOpen`
https://github.com/PeerioTechnologies/peerio-desktop/blob/fd355f6ee695856e585a0b68d03f2a6ada69916f/src/stores/ui-store.js#L21

src/ui/chat/ChatView.jsx
```diff
@@ -197,6 +200,7 @@ class ChatView extends React.Component {
                     }
                     {this.sidebar}
                 </div>
+                {chat.leaving ? <FullCoverLoader show /> : null}
```
anri-asaturov: 👍 😄

src/ui/chat/components/NoChatSelected.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+    render() {
+        if (chatStore.loading) return null;
+        return (
+            <div className="zero-message new-dm create-new-chat">
+                <div style={{
+                    marginTop: '168px',
+                    maxWidth: '600px'
+                }} >
+                    <h2 className="display-2">No chats? <strong>No problem!</strong></h2>
```
anri-asaturov: localization

seavan: this is n mvp thing, it'll get its own proper design at some point.

are you sure you want to add copy which is probably wrong?

anri-asaturov: ok

PR #66, opened 2017-10-16
Fix remembering window state

dchest:
Please read complete the checklist, then fill out any sections that are relevant. Omit sections that are not relevant for cleanliness. 

### PR Checklist

- [ ] I have checked with either the repo owner or person who previously worked on this feature to ensure correctness

https://peerio.tpondemand.com/entity/6689

### Test plan

Eren tested it :)

------


src/main-process/state-persistance.js
```diff
@@ -5,12 +5,14 @@ function saveWindowState(state) {
...
 
 function getSavedWindowState() {
-    const defaultState = {
-        width: 1024,
-        height: 728
-    };
     return TinyDb.system.getValue('windowState')
-        .then(savedState => Object.assign(defaultState, savedState || {}));
+        .then(savedState => ({
+            x: savedState.x,
```
anri-asaturov: will throw null ref exception on clean (not tiny db yet) machines?

src/main.js
```diff
@@ -40,30 +40,65 @@ app.on('ready', () => {
     app.setAppUserModelId(config.appId);
 
     getSavedWindowState()
-        .then(state => {
-            const winConfig = Object.assign(state, {
+        .then(windowState => {
+            const winConfig = Object.assign({
                 show: false,
-                center: true,
+                center: windowState.x == null,
```
anri-asaturov: what about app changing screens? If i used peerio on a larger screen, closed it and then reopened with only my smaller default screen open?

src/main-process/state-persistance.js
dchest: True, fixing

src/main.js
dchest: Hmm, maybe. So, we shouldn't remember coordinates and always center it?

anri-asaturov: @dchest I'm not sure, let's create a test task for @ebilg ? and then it is the case, instead of always centering we could try detecting screen configuration/resolution change or overflow?

src/main-process/state-persistance.js
dchest: Fixed

src/main.js
dchest: I restored centering. Let's merge this for now, and then I'll add tasks for investigating what happens with different screen sizes, and also there's some issue with Ubuntu having a slightly different height (I suspect Electron doesn't account for its title bar size)?

PR #67, opened 2017-10-16
Release overrides

dchest:
There's a new directory `release-overrides` (I'm not attached to the name, tell me if I need to change it), which contains overrides to be applied to the file tree before making a production release (that is, 'peerio-release-builder` will apply them before building or before applying overrides from whitelabel repos).

Files inside this directory will ovewrite files inside the root project directory recursively. For example,`release-overrides/src/config-base.js` will overwrite `src/config-base.js`.

NOTE: currently `release-overrides/src/config-base.js` is the same as `src/config-base.js`, the dev/release separation will be applied to it later as needed.

### PR Checklist

- [ ] I have checked with either the repo owner or person who previously worked on this feature to ensure correctness

### Test plan

* Change something in `release-overrides/src/config-base.js` (e.g. `socketServerUrl`).
* Make a release build.
* Verify that changes in config are applied.

------


anri-asaturov:
TBH having to maintain entire config file in sync feels harder than just passing socket server url to a builder via cli params.
Could you maybe do this release overrides in a way they actually work with json file to override only what's needed?

dchest:
Agree, this sounds better. What do you think about specifying Peerio configuration in app/package.json?

```
{
   "peerio": {
         appId: "...",
         "socketServerUrl": "...
   },
    ...
 }
```

We already use this approach with the updater, and whitelabels override it.

dchest:
Then we'll have `json-overrides` in the main repo with overridden parameters for `app/package.json`

anri-asaturov:
Yup, sounds great, 'json-overrides' as white-label machinery analogue for peerio releases, and additional mechanics to replace single parameters.

dchest:
Yeah. BTW, the mechanics are already there -- it uses lodash/merge, which merges JSONs recursively: https://github.com/PeerioTechnologies/peerio-desktop-staging/blob/master/json-overrides.json

PR #68, opened 2017-10-16
WIP inline image popup on click

suprko:
Markup is done for the inline image popup. The file action buttons (download/share/delete) don't work at the moment.


src/ui/chat/components/InlineFiles.jsx
```diff
@@ -103,30 +175,30 @@ class InlineFile extends React.Component {
...
+                                }
+                            </div>
+                            : <Button className="display-this-image" onClick={this.forceDownload}>
+                                {t('button_displayThisImage')}
+                            </Button>
+
+                        }
+                    </div>
+                }
+                {this.imagePopup}
```
anri-asaturov: It's very inefficient to render separate dialog for every file, ether create one dialog in parent component (chat)
Or do not render the dialog(or at least it's contents) when it's not supposed to be visible

suprko: If instead of `<Dialog active={this.imagePopupVisible}...` I move the bool to `{this.imagePopupVisible && this.imagePopup}`, this would do your 2nd suggestion, right?

anri-asaturov: yes

src/ui/chat/components/InlineFiles.jsx
```diff
@@ -39,33 +37,13 @@ class InlineFile extends React.Component {
...
-    };
-
-    share = () => {
-        fileStore.clearSelection();
-        this.props.file.selected = true;
-        window.router.push('/app/sharefiles');
-    };
+        // need this short timeout to allow Dialog fade-in transition, otherwise it gets squashed
+        setTimeout(() => {
+            this.imagePopupVisible = true;
```
anri-asaturov: It's fine for now, but there's a right way to do it via react `ref`, ask me in slack if you don't understand what i mean

PR #84, opened 2017-10-23
fix: only expand automatically first time. fix delete option in full …

seavan:


src/ui/chat/components/InlineFiles.jsx
```diff
@@ -206,7 +206,8 @@ class InlineFile extends React.Component {
         if (file.isImage && uiStore.prefs.peerioContentEnabled && uiStore.prefs.peerioContentConsented) {
             setTimeout(() => {
                 !file.tmpCached && file.tryToCacheTemporarily();
-                when(() => !file.downloading, () => { this.isExpanded = true; });
+                (this.isExpanded === undefined)
```
anri-asaturov: Is there a way not to do it in render()? We are starting to make a pattern from this ugliness..

PR #92, opened 2017-10-26
Feat files icons

suprko:
### PR Checklist

File Icon images! Currently only exists on FileLine component, used in Files view.

### SDK notes
Depends on https://github.com/PeerioTechnologies/peerio-icebear/pull/52


src/ui/files/components/FileLine.jsx
```diff
@@ -14,6 +14,83 @@ const moment = require('moment');
...
     @observable showActions = false;
+    @observable isImage = false;
+
+    @computed get fileIconType() {
+        let extension = this.props.file.name.toLocaleLowerCase().match(/\.\w+$/);
+        extension = extension ? extension[0].substring(1) : '';
+
+        let fileType = '';
+
+        switch (extension) {
```
anri-asaturov: Extract this to some separate helper file please, and convert switch to an object that looks like 
```{
  bmp: 'img',
  git: 'img'
}```
Maybe generate that object if you don't want to be a monkey 🐒 
Something like
```const obj = {}
['bmp', 'png','etc'].forEach(ext=>ob[ext]='img')
```

src/ui/files/components/FileLine.jsx
```diff
@@ -14,6 +14,83 @@ const moment = require('moment');
 @observer
 class FileLine extends React.Component {
     @observable showActions = false;
+    @observable isImage = false;
```
anri-asaturov: what is this flag doing? Btw File object has isImage flag.

suprko:
@anri-asaturov 

* file type check moved to `file.js` helper
* file type check changed from switch to object
* icons extracted into new FileSpriteIcon component

src/helpers/file.js
```diff
@@ -99,4 +99,26 @@ function getListOfFiles(paths) {
...
+    ai: 'ai',
+    psd: 'psd'
+};
+function createFileType(extension, type) { fileType[extension] = type; }
+['bmp', 'gif', 'jpeg', 'jpg', 'png', 'svg', 'tif', 'tiff'].forEach((ext) => createFileType(ext, 'img'));
+['aif', 'aiff', 'flac', 'm4a', 'mp3', 'ogg', 'opus', 'wav'].forEach((ext) => createFileType(ext, 'audio'));
+['avi', 'flv', 'm4v', 'mov', 'mp4', 'mpeg', 'mpg', 'wbm', 'wmv'].forEach((ext) => createFileType(ext, 'video'));
+['7z', 'gz', 'rar', 'zip', 'zipx'].forEach((ext) => createFileType(ext, 'zip'));
+['doc', 'docx'].forEach((ext) => createFileType(ext, 'word'));
+['xls', 'xlsx'].forEach((ext) => createFileType(ext, 'xls'));
```
anri-asaturov: What about apple spreadhsheets and docs? Open office? )
Maybe there's an open database we can use somewhere?
I don't insist doing it now, but adding it to TP.

suprko: As a LibreOffice user I agree :D

The thing is that the icons themselves are for MS Office. e.g. .doc icon is a blue "W". I would prefer more generic icons -- document, spreadsheet, etc.

src/ui/files/components/FileLine.jsx
```diff
@@ -79,7 +82,12 @@ class FileLine extends React.Component {
                             onCancel={this.cancelUploadOrDownload} />
                         : <Checkbox disabled={!file.readyForDownload} checked={file.selected}
                             onChange={this.toggleChecked} />
-                    }</td>
+                    }
+                </td>
+
+                <td className="file-icon">
+                    <FileSpriteIcon type={getFileType(this.props.file.name)} size="medium" />
```
anri-asaturov: let's make 'type' a computed propery of File object in icebear, this way desktop and mobile can reuse it and we'll have to call getFileType only once per file.
Here it is https://github.com/PeerioTechnologies/peerio-icebear/blob/dev/src/models/files/file.js
While you're there maybe rename `File#isImage` to `File#isInlineableImage` or find better name.

anri-asaturov: Ah and helpers with type mapping goes to sdk of course.

src/ui/shared-components/FileSpriteIcon.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+class FileSpriteIcon extends React.Component {
+    /* props
+        size: medium, large
+        type: img, audio, video, txt, zip, pdf, ai, psd, word, xls, ppt, other
+    */
+
+    fileType = `ic-file-${this.props.type}-${this.props.size}`;
+
+    @computed get pxResolution() {
```
anri-asaturov: you don't need a computed here, your property never changes, but even if it does - not worth creating a computed value. Btw props are automatically observable.

anri-asaturov: (jsut remove @computed decorator)

src/ui/shared-components/FileSpriteIcon.jsx
```diff
@@ -0,0 +1,27 @@
...
+const { computed } = require('mobx');
+const css = require('classnames');
+
+class FileSpriteIcon extends React.Component {
+    /* props
+        size: medium, large
+        type: img, audio, video, txt, zip, pdf, ai, psd, word, xls, ppt, other
+    */
+
+    fileType = `ic-file-${this.props.type}-${this.props.size}`;
```
anri-asaturov: Did you test this? It's not gonna work bcs it's executed only in component constructor. props might be already available, i don't remember, but changing type and size won't do anything.
It's ok to calculate this in 'render'

suprko: I tested it, it does work! `type` and `size` never change, they are provided wherever `<FileSpriteIcon>` is used

PR #98, opened 2017-10-30
fix: "git+ssh" in package.json breaks npm install for folks w/o ssh keys

lostfictions:


lostfictions:
oops, ok

lostfictions:
yeah, any repo-ish convention is fine, it's not platform-specific. if Samvel was using a different git client that defaulted to https instead of ssh he'd have had the same problem i think

lostfictions:
you can even use just `RepoOwner/repo-name#branch`, no need for `github:` afaik

anri-asaturov:
ok, i'll try it

PR #99, opened 2017-10-31
chore: stylelint embetterment

lostfictions:
This PR changes the stylelint `property-ordering` rule, as discussed in Slack today. It also changes the precommit hook to lint all styles, not just those in `src/styles`.

Some notes on the changed rule:
- Prefixed versions of properties are [always required to be listed before the unprefixed version](https://github.com/hudochenkov/stylelint-order/blob/master/rules/properties-order/README.md), so there's no need to write them out explicitly; I've removed all of those.
- There were some rules for frameworks we're not using (lostgrid) which I also removed.

Things to keep in mind when writing new CSS:
- Unrecognized properties (or ones not listed in the rule) have to go at the top of the block.
- Otherwise, I tried to sort things in broad categories with most important stuff at the top. I couldn't add comments because it's a json file, so I made dummy properties as comments for dividing the categories.

The PR also commits the results of running stylelint in fix mode (and some other fixes from files that weren't previously linted that would fail the hook or CI) which is why the diff is so big. Looking over the PR by commit should be more manageable.

### PR Checklist


suprko:
At first glance I like it. I'll have to see how it feels in practice. `color` is always weird because it can affect many things.

lostfictions:
@suprko Agreed, maybe appearance rules should go before typography/flow/etc?

lostfictions:
Hey Anri, sorry to reopen this, but I think Lucas is right -- the color/appearance stuff should be bumped up, it's confusing for it to be near the bottom. Now that this is already merged, easiest if I just submit a new PR from the same branch...?

anri-asaturov:
Whatever works for you.

PR #101, opened 2017-11-02
fix: remove 2fa autosubmit

suprko:
### PR Checklist

2FA dialog no longer autosubmits after the 6th character has been entered. "Submit" button needs to be pressed manually.
https://peerio.tpondemand.com/entity/7178

### Test plan
So far only tested by Lucas. This change affects
* 2FA dialog on login
* 2FA dialog when choosing to disable 2FA

This change *does not affect*
* 2FA input in Security Settings, when deciding to enable 2FA. <- this one still autosubmits


src/ui/shared-components/TwoFADialog.jsx
```diff
@@ -11,11 +11,30 @@ const appControl = require('~/helpers/app-control');
 @observer
 class TwoFADialog extends React.Component {
     @observable totpCode = '';
+    @observable readyToSubmit = false;
```
lostfictions: `readyToSubmit` should be a `@computed` value, I think -- this is a really good way to prevent bugs down the line resulting from the introduction of unneeded state, and it's one of the big advantages of using something like MobX. Did @anri-asaturov discuss this when you guys did a MobX overview?

src/ui/shared-components/TwoFADialog.jsx
```diff
@@ -58,17 +72,6 @@ class TwoFADialog extends React.Component {
         if (clientApp.active2FARequest) ref.focus();
     }
 
-    onTOTPCodeChange = ev => {
```
lostfictions: Eesh, looks like `onTOTPCodeChange` was getting redeclared in the body of the class? @anri-asaturov know if there's a lint to forbid this? (Good that you caught this, Lucas)

src/ui/shared-components/TwoFADialog.jsx
```diff
@@ -11,11 +11,25 @@ const appControl = require('~/helpers/app-control');
 @observer
 class TwoFADialog extends React.Component {
     @observable totpCode = '';
+    @observable readyToSubmit = false;
 
     onTOTPCodeChange = ev => {
         this.totpCode = ev.target.value;
```
lostfictions: `this.totpCode` is an observable that you're changing in this callback, so the callback should be decorated with `@action` (or simpler, switched to a non-arrow function with `@action.bound` like you have with `submitCode()` below.)

src/ui/shared-components/TwoFADialog.jsx
```diff
@@ -77,7 +80,12 @@ class TwoFADialog extends React.Component {
         const req = clientApp.active2FARequest;
         let actions;
         if (req) {
-            actions = [{ label: t('button_cancel'), onClick: this.cancel }];
+            actions = [
+                { label: t('button_cancel'), onClick: this.cancel },
+                { label: t('button_submit'),
+                    onClick: this.submitCode,
```
lostfictions: indentation

src/static/locales/en.json
```diff
@@ -59,7 +60,6 @@
   "title_welcomeBack": "Welcome back,",
   "title_signupStep1": "Signup (1 of 2)",
   "title_signupStep2": "Signup (2 of 2)",
-  "title_signupAvatar": "Add photo (optional)",
```
lostfictions: Why was this removed?

src/ui/shared-components/TwoFADialog.jsx
```diff
@@ -11,11 +11,25 @@ const appControl = require('~/helpers/app-control');
...
         this.totpCode = ev.target.value;
+        const res = validateCode(this.totpCode);
+        if (res.readyToSubmit) {
+            this.readyToSubmit = true;
+        } else {
+            this.readyToSubmit = false;
+        }
     };
 
+    @action.bound
```
anri-asaturov: Let's maybe not do new line after decorators on class members. @six?

src/ui/shared-components/TwoFADialog.jsx
lostfictions: (This is incidentally one of the pitfalls of using `callback = () => { ... ` class fields instead of `callback()` methods, and one reason why something like `@action.bound` is often preferable)

src/ui/shared-components/TwoFADialog.jsx
lostfictions: I prefer each decorators on its own line, personally -- when you're scanning a file looking for a method name it makes it easier to find.

anri-asaturov: it makes files so much longer...

anri-asaturov: Do you know if eslint or tslint can deal with this?

src/ui/shared-components/TwoFADialog.jsx
suprko: So in addition to `onTOTPCodeChange` getting declared twice, there was stuff inside the function that should have thrown errors but wasn't, e.g. `this.totpCodeError` and `this.totpCodeValidating` don't exist...

src/static/locales/en.json
suprko: Removed from peerio-copy by Karim, I guess it's not being used anymore: https://github.com/PeerioTechnologies/peerio-copy/commit/dfa2379a9cc7b9e71f03af8a5d5b99ad8a19775f

src/ui/shared-components/TwoFADialog.jsx
anri-asaturov: redeclaration of class members? probably no, no rule. There's only this https://eslint.org/docs/rules/no-redeclare
and another one that prevents reassigning function args

lostfictions: Yeah, when you redeclare a field like this it's basically up to Babel to figure out what to do with it -- it should probably be exploding and throwing an error on compile but it wasn't, so I guess maybe the first declaration won and this one was just dead code...?

src/static/locales/en.json
lostfictions: :+1: (Maybe for clarity, when you update from `peerio-copy` it's worth organizing as a separate commit)

src/ui/shared-components/TwoFADialog.jsx
anri-asaturov: It's weird that eslint is not bothered by this

anri-asaturov:
@ebilg waiting for your review

src/ui/shared-components/TwoFADialog.jsx
suprko: Sry I'm not seeing where the indentation is wrong? This is what eslint corrected it to

anri-asaturov: second property is indented, i think the reason is that first property is on the same like with `{`.
Try like this
```
{
   label: ...,
   onClick:...

Comment (suprko) [reply src/ui/shared-components/TwoFADialog.jsx]: k fixed

File: src/ui/shared-components/TwoFADialog.jsx [hunk 7]
```diff
@@ -11,11 +11,20 @@ const appControl = require('~/helpers/app-control');
 @observer
 class TwoFADialog extends React.Component {
     @observable totpCode = '';
+    @computed get readyToSubmit() {
+        return validateCode(this.totpCode).readyToSubmit;
+    }
 
     onTOTPCodeChange = ev => {
```
lostfictions: `onTOTPCodeChange` should still ideally be `@action.bound` (like Anri said, it's not 100% necessary with our current setup, but it's a good habit to get used to moving forward and makes both [performance and tooling](https://mobx.js.org/refguide/action.html) better)

ebilg:
commented on TP, without autosubmit, it would be nice to be able to press enter to submit

suprko:
@ebilg input now responds to the enter key added. let us know if all's good.

PR #102, opened 2017-11-04
fix: chat autoscrolls on new reciepts

anri-asaturov:
#7345

Please read complete the checklist, then fill out any sections that are relevant. Omit sections that are not relevant for cleanliness. 

### PR Checklist

### Test plan

*Examples of things to note here:*
*This PR makes changes to the following shared components: ____ please test the component in all its contexts.*
*Here is some behaviour that could have regressed: ____.*
*Aside from QA and the repo owner, the following people should check this PR: _____ for reasons _____*
*To test this feature, use the following external tools/process: ____*

### SDK notes

*link to related PR in peerio-icebear*

### Gotchas/additional notes for reviewer

------


anri-asaturov:
test test. Will this sync with TP?

src/ui/chat/components/MessageList.jsx
```diff
@@ -201,11 +201,25 @@ class MessageList extends React.Component {
...
+    lastRenderedChatId = null;
     render() {
         if (!chatStore.activeChat) return null;
+        // we want to autoscroll only if there are new messages
+        if (this.lastRenderedChatId !== chatStore.activeChat.id) {
+            this.lastRenderedMessageId = null;
+            this.lastRenderedChatId = chatStore.activeChat.id;
+        }
+        const messages = chatStore.activeChat.messages;
+        if (messages.length) {
```
lostfictions: I had to think about this for a sec -- by `if (messages.length)` you meant `messages.length > 0` and not that it's possibly null, right? In general I'm in favour of the expression being close to your intent if possible, even if we can shave a few characters off with falsiness :)

src/ui/chat/components/MessageList.jsx
```diff
@@ -201,11 +201,25 @@ class MessageList extends React.Component {
...
+            if (this.lastRenderedMessageId !== messages[messages.length - 1].id) {
+                this.lastRenderedMessageId = messages[messages.length - 1];
+                setTimeout(this.scrollToBottomIfSticky, 400);
+            }
+        } else {
+            this.lastRenderedMessageId = null;
+        }
```
lostfictions: I get that this is linked to rerenders and putting it here is maybe more explicit about that than if it were in an autorun/reaction, but I'm not really a fan of having assignments or side effects in the render function. (And if the logic for checking scrolling gets more complicated later and derefs an observable unrelated to the render logic, it could trigger unneeded rerenders.)

What about having this in the [`componentWillReact` hook](https://mobx.js.org/refguide/observer-component.html) MobX provides?

src/ui/chat/components/MessageList.jsx
anri-asaturov: I don't find this too fancy tbh, normal js stuff

anri-asaturov: I get it you probably are not used to it using TS mostly?

src/ui/chat/components/MessageList.jsx
anri-asaturov: Yes, that's a good idea i didn't know of this hook

anri-asaturov: Updated.
Actually used `componentDidUpdate` instead triggering scroll before changes are rendered feels wrong, even with timeout. Also removed some reactions that became redundant and fixed a bug, thanks for review.

src/ui/chat/components/MessageList.jsx
dchest: Yeah, I also prefer clearer expressions, as they are easier to read compared to "normal js stuff" :) But it's fine.

lostfictions:
Yeah, looks good to me, thanks for the changes!

ebilg:
bug fixed, behaviour looks correct to me

PR #103, opened 2017-11-04
fix(ux): show exact timestamp for files older then 21 hour

anri-asaturov:
https://peerio.tpondemand.com/entity/7344


src/ui/files/components/FileLine.jsx
```diff
@@ -56,18 +56,28 @@ class FileLine extends React.Component {
...
+    }
+
     render() {
         const file = this.props.file;
         if (!file.show) return null;
         // minuteClock.now is never null, but this will subscribe us to clock events to re-render relative timesamp
         if (!uiStore.minuteClock.now) return null;
-
+        // just to improve performance of this.formatDate(..)
+        this.renderTimestamp = Date.now();
```
lostfictions: Wouldn't it be even more performant (and creating less state) to just start using the value of `uiStore.minuteClock.now` that's being ignored two lines up?

anri-asaturov: Minute clock updates every minute, meaning it has the same value for entire minute and we need current time.

anri-asaturov: Although probably 1 minute lag won't give us much difference.

lostfictions: Hm, not sure I understand the difference... You mean just for the first minute after files were created, so we can see "xx seconds ago" instead of "now" or "x minutes ago"?

anri-asaturov: The difference is only in the last minute of 21h interval, if we use minute clock, we can be late to the 21h limit by 1 minute, but that's not a problem at all in this case, no one will notice :-D

lostfictions: But `reportObserved()` is already going to get invoked on every render because of the invocation of `minuteClock.now` up there...? Might as well use it?

If you want to be explicit (and not invoke the getter twice) you could grab the value in the render function here and explicitly pass it to both methods?

anri-asaturov: Yeah, i like it

anri-asaturov: Can justify the observable dereference now.

anri-asaturov:
updated

PR #104, opened 2017-11-07
feat: proper file size warning. rename and unify of oversized

seavan:
refactor and actually make the warnings work.


src/ui/chat/components/InlineFiles.jsx
```diff
@@ -37,8 +39,28 @@ class InlineFile extends React.Component {
...
+        this._fileReaction = reaction(() => fileStore.getById(this.props.id), file => {
+            if (!file) {
+                this.startTimer();
+                return;
+            }
+            const { peerioContentEnabled, peerioContentConsented } = uiStore.prefs;
+            if (file.isImage && peerioContentEnabled && peerioContentConsented) {
+                !file.tmpCached && file.tryToCacheTemporarily();
+                (this.isExpanded === undefined)
+                    && when(() => !file.downloading, () => { this.isExpanded = true; });
```
anri-asaturov: Come on, man, don't hate if so much it has done nothing to you.

src/ui/chat/components/InlineFiles.jsx
```diff
@@ -37,8 +39,28 @@ class InlineFile extends React.Component {
...
+            }
+            this._pathReaction = reaction(() => file.tmpCachePath, () => {
+                this.errorLoading = false;
```
anri-asaturov: could you comment on why are we doing this please (in the code)

seavan: actually I see lucas's another branch, it's solved there. removing.

anri-asaturov:
Change sdk version to 3.0.0 and request a review from Eren afterwards, please.

suprko:
Made a tiny spacing fix, otherwise looks/works fine for me

seavan:
re-review re-requested re: if

src/ui/chat/components/InlineFiles.jsx
```diff
@@ -37,8 +39,25 @@ class InlineFile extends React.Component {
...
+        this._fileReaction = reaction(() => fileStore.getById(this.props.id), file => {
+            if (!file) {
+                this.startTimer();
+                return;
+            }
+            const { peerioContentEnabled, peerioContentConsented } = uiStore.prefs;
+            if (file.isImage && peerioContentEnabled && peerioContentConsented) {
+                if (!file.tmpCached) file.tryToCacheTemporarily();
+                if (this.isExpanded === undefined) {
+                    when(() => !file.downloading, () => { this.isExpanded = true; });
```
anri-asaturov: Technically file will stop downloading when there's an error like disconnection or something. 
Maybe u wanted to use `file.tmpCached`?

src/ui/chat/components/InlineFiles.jsx
```diff
@@ -204,17 +221,20 @@ class InlineFile extends React.Component {
         );
     }
 
+    renderOversizeCutoffWarning() {
+        return (
+            <div className="image-over-limit-warning">
+                <T k="title_imageTooBigCutoff" className="text" >
+                    {{ size: util.formatBytes(config.chat.inlineImageSizeLimitCutoff) }}
```
anri-asaturov: let's cache this  object `{size:util.formatBytes(config.chat.inlineImageSizeLimitCutoff)}` it never changes

src/ui/chat/components/InlineFiles.jsx
```diff
@@ -37,8 +39,25 @@ class InlineFile extends React.Component {
         }, 30000);
     };
 
+    componentWillMount() {
+        this._fileReaction = reaction(() => fileStore.getById(this.props.id), file => {
```
anri-asaturov: Please add comments describing what does this do for the sake of other devs

src/ui/chat/components/InlineFiles.jsx
anri-asaturov: Or you want to expand it to show the error?

anri-asaturov:
@ebilg it's your turn to critique Samvel's PR :-D

anri-asaturov:
Both @seavan and @ebilg this PR waits for you.

src/ui/chat/components/InlineFiles.jsx
lostfictions: Is there a reason we use reactions instead of autorun like this? I find autorun simpler and clearer.

anri-asaturov: Generally, we always try to use reactions for more granular and safe subscriptions. This particular thing I don't like too much but don't know how to make it better yet.

seavan: autorun can be very scary with its side effects, especially how every other thing in icebear is an observable.

lostfictions: But often that's behaviour you want (for example, in this case, if the user consents or changes their preferences, that change should apply to other files in the message view too -- they shouldn't have to reload to see the effect.)

seavan: not really, we don't want to trigger one hundred re-renders because user updated a setting, that may become a major performance issue.

rather that making sure that every autorun's body does not create a ton of subscriptions, it's easier and more manageable to single out the things which you do want to react too.

also this code is ugly either way, but as Anri said, no idea yet how to make it better. it should probably be moved to icebear for most part.

lostfictions: In fact, you're convincing me more and more this particular case should be an autorun (and it should check if the timer is already active) :laughing:

seavan: autorun is evil. if you have side effect in autorun body which updates one of the observables in the autorun body, you got yourself a nasty loop. is that convincing you - making more deadloop prone code? you have interesting tastes :D

anri-asaturov: @lostfictions In this particular case (which should be burned with fire soon, since we moved to cassandra and can afford to make queries to check file status) we want to react to file appearing in observable map (that's what `getById()` does) and not to `files` collection changes or anything else getting caught into autorun.

src/ui/chat/components/InlineFiles.jsx
```diff
@@ -37,8 +41,27 @@ class InlineFile extends React.Component {
...
+            if (!file) {
+                this.startTimer();
+                return;
+            }
+            const { peerioContentEnabled, peerioContentConsented } = uiStore.prefs;
+            if (file.isImage && peerioContentEnabled && peerioContentConsented) {
+                if (!file.tmpCached) file.tryToCacheTemporarily();
+                if (this.isExpanded === undefined) {
+                    when(() => file.tmpCached, () => { this.isExpanded = true; });
```
seavan: @anri-asaturov tmpCached is better, you're right

src/ui/chat/components/InlineFiles.jsx
```diff
@@ -11,11 +11,15 @@ const routerStore = require('~/stores/router-store');
...
 const FileActions = require('~/ui/files/components/FileActions');
 const { getAttributeInParentChain } = require('~/helpers/dom');
+const config = require('~/config');
+const { util } = require('~/icebear');
 
 const ALL_IMAGES = 'all_images';
 const UNDER_LIMIT_ONLY = 'under_limit_only';
 const DISABLED = 'disabled';
 
+const cutoffSizeFormatted = util.formatBytes(config.chat.inlineImageSizeLimitCutoff);
```
seavan: @anri-asaturov caching!

seavan:
@anri-asaturov review requested

src/ui/chat/components/InlineFiles.jsx
lostfictions: (this is totally a nit, but if we're going to the trouble of caching, couldn't the whole object be cached so this just becomes `const cutoffSizeFormatted = { size: util.formatBytes(config.chat.inlineImageSizeLimitCutoff) }` and you pass that to the `<T>`?)

lostfictions: (on the other hand it's basically free to build objects in V8 so don't let me make this PR even more torturous :laughing: )

anri-asaturov: @lostfictions that was my original suggestion ;) I just seem to have confused Sam

seavan:
@ebilg @anri-asaturov fixed in client.

seavan:
@anri-asaturov @lostfictions round 2

anri-asaturov:
@seavan Eren is still testing right?

anri-asaturov:
@seavan conflicts

seavan:
@anri-asaturov @ebilg merged and updated

src/config.js
```diff
@@ -42,6 +44,14 @@ cfg.upload.uploadBufferSize = 1024 * 1024 * 3;
 // --- PERFORMANCE
 cfg.chat.maxInitialChats = 20;
 
+// --- IMAGE DISPLAY LIMITS
+cfg.chat.inlineImageSizeLimit = 10 * 1024 * 1024;
+cfg.chat.inlineImageSizeLimitCutoff = 30 * 1024 * 1024;
+
+// this will go to icebear on next iteration
+cfg.chat.inlineImageSizeFormatted = { size: util.formatBytes(cfg.chat.inlineImageSizeLimit) };
+cfg.chat.inlineImageSizeLimitCutoffFormatted = { size: util.formatBytes(cfg.chat.inlineImageSizeLimitCutoff) };
```
anri-asaturov: this is not a part of config

src/ui/chat/components/InlineFiles.jsx
```diff
@@ -260,25 +281,29 @@ class InlineFile extends React.Component {
...
-                                    }
+                                                onClick={this.imageClick} />)}
+                                    {!file.tmpCached && !file.downloading && file.isOverInlineSizeLimit
+                                        && !file.isOversizeCutoff
+                                            && this.renderOversizeWarning()}
+                                    {file.isOversizeCutoff
+                                            && this.renderOversizeCutoffWarning()}
+                                    {file.downloading && !file.tmpCached &&
+                                        <img src="./static/img/bg-pattern.png" alt="" />}
+                                    {file.cachingFailed ? <span>{t('error_downloadFailed')}</span> : null}
```
anri-asaturov: All this conditions are pretty hard to read and easy to break, i would extract it to plain js code right before returning component.
something like 
```
render(){
...
  let component;
  if(file.tmpCached){
     component = ...
  } else if(file.downloading){
     component = ...
  }


 return <>  {component} </>
}```

seavan:
@anri-asaturov bump

PR #106, opened 2017-11-08
feat(messages): new ProseMirror-based message input

lostfictions:
First, apologies for the size of this PR. I realize something like this is difficult to review in terms of a single unit of change, and moving forward if I realize I'm working on something of this scale I'll at least try to split each unit of functionality into its own sub-branch off the feature branch for clarity. (Some of the functionality here probably could've been PRed and merged into dev without disrupting anything until we were ready to "hit the switch", too.) I know it's hard to look over in Github like this, but if anyone has time to pull the branch and step through the commits (though I know there's still a lot of those too) I'd really appreciate a critical eye.

Anyway this PR switches the chat message input to the new ProseMirror-backed implementation.

**New/changed functionality:**
- Write bold text with `ctrl/cmd-b` -- either by selecting text and toggling, or toggling and then typing
- Write italic text with `ctrl/cmd-i`, as above
- Write strikethrough with `ctrl/cmd-5`, as above

- Rich text (HTML) can be copied from elsewhere and pasted into the input and any applicable formatting will be preserved. (A task for a "clear all formatting" hotkey is already in TP.)

- Mentions
  - Mentions are displayed inline while composing like they would be in the chat list, with a background colour
  - Mention suggestions are offered based on participants in the room (same as the current functionality). A user can type out a mention for which there's no suggestion and it will still work (useful if you want to mention someone in a DM, or someone who's not currently in a room).

- Emoji
  - Emoji are displayed inline; starting to typing a shortname like `:dog:` offers an autocomplete after two characters (or you can type the whole shortcode manually and it'll parse when you hit the closing `:`).
  - The emoji autocomplete is naive and just filters against the start of every shortname with no scoring of suggestions; improving the quality of matches is potential future work.
  - Copying and pasting emoji (within the app or from an external source) now works as expected.

- Incidentally, autocomplete suggestions for both emoji and mentions can now be confirmed with either `tab` or `enter`.

- URLs
  - Unlike the other entities, URLs are not displayed inline while composing a message -- they're only parsed when you send the message.
  - URL parsing logic should be as robust as the current implementation -- it uses the same parsing core, which was the most robust I could find.
  - URLs are marked up in the rich text payload but aren't allowed to point to a different location than their text (see [discussion in chat-schema.jsx](https://github.com/PeerioTechnologies/peerio-desktop/blob/abc4bfa85862350f3e88c6290a44e4af671742e6/src/helpers/chat/prosemirror/chat-schema.jsx#L74-L88)) so we still end up doing some normalization of the final href on the receiving side. (This has no practical impact and works the same as it does in the current implementation.)

This implementation also produces a plaintext version of the message which is sent alongside the rich text in the message keg payload. It's used for backwards-compatibility (ie., rich text messages are readable in plaintext on older clients) but also for system notifications, via Icebear.

--

To do before merge:
- [x] Update deps (ProseMirror has hit 1.0, but this PR uses the RC so it shouldn't require major updates)
- [x] Add localization strings if/where necessary (needs to be looked over)

Might be able to wait until after merge:
- Figure out a story for forwards-compatibility with additional rich-text formatting. ie, graceful degradation when new rich text formatting types are added to the ProseMirror schema in new versions of the client/SDK and aren't recognized on older clients. (Ditto if formatting types are removed/modified for whatever reason, too.)

  Currently if there's a parse error in the schema the message will downgrade to displaying plaintext, which is a bit crude but might be enough for now.

- Update Icebear to bring underlying behaviour in line with new message handling
  - Migrate the message schema and any other shared functionality into Icebear (may also be necessary for upcoming story to support display of rich text on mobile)
  - Mention handling in Icebear should use the same logic as the chat schema (and not duplicate parsing work) to avoid, for example, a notification but no highlight (example of this kind of false positive: `me@six.com` would ping `@si` with the old parsing, but with the new message format the former would not be highlighted)
  - URL unfurling in Icebear should use the same logic as the chat schema (and ideally not duplicate parsing work). Currently it uses `url-regex` and has discrepancies vs. linkifying, even in the current release client (eg. it tries -- and fails -- to unfurl schemaless URIs like //google.com even if the latter doesn't display as a link.)

---

### PR Checklist

[TP#825](https://peerio.tpondemand.com/entity/825) (and bugs/stories in [TP#650](https://peerio.tpondemand.com/entity/650) more generally)

### Test plan

- All messages sent in the old client should display identically in the new client.
- All messages sent in the new client should display in the old client (without rich text formatting). Emoji, mentions, and URLs should still parse/display in the old client.
- Any overlapping combination of emoji, mentions, bold, italics, URLs, etc. should display correctly
- Lots of markup/entities and multiple URLs in a message (consecutive and/or spaced out) should work
- Newlines (shift-enter) should behave as they do now and shouldn't mess anything up
- System notifications should not display any anomalies and should display emoji correctly
- Autocomplete for mentions and emoji shouldn't get in the way of any input unexpectedly (eg. swallow input when a user might be trying to type)
- Undo/redo should always behave as expected
- It should be possible to send something that looks like a partial emoji shortname (eg `:to`) without it being coerced into an emoji
- It shouldn't be possible to get an emoji in the message input box that _isn't_ displayed as emojione
  - System emoji keyboards should work
  - Copying and pasting emoji from any source should work (except images of emoji, obviously)
  - Inputting composite emoji like the ["family" emoji](https://emojipedia.org/families/) should work regardless of input method/paste source
- It might be possible to create a "mention-like" thing (an @ followed by a valid username pattern and preceded by whitespace or the beginning of the line) without it displaying as such in the input box if you get tricky and move your cursor around a lot, or if you dismiss the suggestions and hit enter. However, it _shouldn't_ be possible to send such a "mention-like" thing without it turning into a proper mention entity on send.

### SDK notes

Rich text field support added in PeerioTechnologies/peerio-icebear#49 (merged Oct 19)

### Gotchas/additional notes for reviewer

------


src/helpers/chat/prosemirror/chat-schema.jsx
```diff
@@ -0,0 +1,284 @@
...
+                // created ProseMirror node or mark. If we return null, that
+                // means the rule matches but we're not providing any attributes
+                // (whereas 'false' means the rule didn't match.)
+                // @ts-ignore (bad typings)
+                { tag: 'span', getAttrs: node => node.classList.contains('strike') && null },
+                { style: 'text-decoration=line-through' },
+                { tag: 'strike' },
+                { tag: 's' }
+            ],
+            toDOM() { return ['span', { class: 'strike' }]; }
```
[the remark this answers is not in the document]
lostfictions: My reasoning was for design -- if they decided the style needed to be change, they'd have an unopinionated base instead of having to override UA styles. But you're right, we already use `em` and `strong`, so it makes sense for consistency.

src/helpers/chat/prosemirror/emoji-plugin.js
```diff
@@ -0,0 +1,82 @@
...
+
+const parser = DOMParser.fromSchema(chatSchema);
+
+function emojiPlugin() {
+    // ProseMirror's handleTextInput prop seems to always work with surrogate
+    // halves, so an IME that inserts an emoji may call it twice or more with
+    // meaningless characters each time.
+    //
+    // As a solution, we accumulate characters with a short debounce, and check
+    // them all on the trailing edge.
```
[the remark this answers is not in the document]
lostfictions: Looks like `handleTextInput` can be invoked in two places, one of which is just the [`keypress` handler](https://github.com/ProseMirror/prosemirror-view/blob/aed41b5ecccf0e77a24f17990f751fcd682e4b55/src/input.js#L95) for the contenteditable view. (The other seems to be something [related to tracking DOM changes](https://github.com/ProseMirror/prosemirror-view/blob/0c909196f2842b15860225e87a26a4521111d4d9/src/domchange.js#L261), so it might potentially be called with more than one char code, but I'm not sure.) It might theoretically be possible for the accumulator to catch non-emoji char codes alongside emoji char codes if a user had some special IME that sent everything in a batch. (But I'm having trouble imagining a tool that would do that, even an accessibility tool.) The failure case is that the emoji get sent as unicode instead of tagged in the markup, which isn't disastrous.

I don't think it makes sense to start/stop the accumulator on individual surrogate halves, because we also have to handle the case of combined emoji (which might be comprised of eight or more surrogate halves, potentially with ZWJs in there too.)

src/ui/chat/components/Message.jsx
```diff
@@ -170,4 +237,49 @@ class Message extends React.Component {
...
+                <div className="message-content">
+                    <div className="message-body">
+                        <p>
+                            <strong>Error rendering message:</strong><br />
+                            {errorData.error.toString()}<br />
+                            <strong>Message:</strong><br />
+                            { // We can't just stringify the message keg, it's not plain data and can be circular
+                                msg ?
+                                    [
+                                        `Sender: ${msg.sender && msg.sender.username}`,
```
lostfictions: Should the error boundary be localized? Ideally no end user will ever see this...

src/ui/chat/components/Message.jsx
```diff
@@ -1,33 +1,103 @@
...
+    }, {}>}
  */
 @observer
 class Message extends React.Component {
+    @observable.shallow errorData = null;
+
+    @action
+    onClickContact(ev) {
+        uiStore.contactDialogUsername = ev.target.attributes['data-username'].value;
+    }
```
anri-asaturov: Let's pass this action in props, do some premature optimisation :)

src/ui/chat/components/Message.jsx
```diff
@@ -61,25 +131,20 @@ class Message extends React.Component {
             </div>
         );
     }
+
+    componentDidCatch(error, info) {
+        runInAction(() => { this.errorData = { error, info }; });
```
anri-asaturov: hmm, marking it as @action doesn't work?

src/ui/chat/components/Message.jsx
```diff
@@ -61,25 +131,20 @@ class Message extends React.Component {
...
-            }
-        */
         const m = this.props.message;
-        // console.log('Rendering message ', m.tempId || m.id);
+
+        if (this.errorData) { return renderError(this.errorData, m); }
+
         const invalidSign = m.signatureError === true;
 
+        const MessageComponent = this.getMessageComponent(m);
```
anri-asaturov: We can cache this, if `message.version` hasn't changed - content is the same.

src/ui/chat/components/Message.jsx
anri-asaturov: Localized as in 'translated'? Yes, let's do it bcs this is pretty much hacking attempts and shouting at people in a language they don't understand is making it worse.

src/ui/chat/components/MessageInputProseMirror.jsx
```diff
@@ -0,0 +1,340 @@
...
+        if (this.editorView) this.editorView.updateState(this.getEmptyState());
+    };
+
+    backupUnsentMessage(chat) {
+        if (!chat || !this.editorView) return;
+        uiStore.unsentMessages[chat.id] = this.editorView.state.doc.toJSON();
+    }
+
+    restoreUnsentMessage(chat) {
+        // TODO: revisit the need for this try/catch? it's a holdover from Quill
```
anri-asaturov: Qull, being not a react component was not in sync with react lifecycle and this function did indeed throw sometimes. In any case there's no harm in keeping the exception handler.

src/ui/chat/components/suggestions/EmojiSuggestions.jsx
```diff
@@ -0,0 +1,55 @@
...
+                .filter(name => name.startsWith(matchData.match[1].toLowerCase()))
+                .map(name => ({ emoji: emojiByAllShortnames[name], name }));
+        },
+        formatter: data => (
+            <span>
+                <img
+                    className="emojione"
+                    alt={data.emoji.characters}
+                    src={`${pngFolder}${data.emoji.unicode}.png`}
+                    style={{ marginRight: 8 }}
```
anri-asaturov: ;)

src/ui/chat/components/suggestions/EmojiSuggestions.jsx
```diff
@@ -0,0 +1,55 @@
...
+            return shortnames
+                .filter(name => name.startsWith(matchData.match[1].toLowerCase()))
+                .map(name => ({ emoji: emojiByAllShortnames[name], name }));
+        },
+        formatter: data => (
+            <span>
+                <img
+                    className="emojione"
+                    alt={data.emoji.characters}
+                    src={`${pngFolder}${data.emoji.unicode}.png`}
```
anri-asaturov: another premature optimization idea - should we pre-create these strings in emoji data object by unicode? It's used in several places.

src/ui/chat/components/suggestions/EmojiSuggestions.jsx
```diff
@@ -0,0 +1,55 @@
...
+const { EditorView } = require('prosemirror-view'); // eslint-disable-line no-unused-vars, (for typechecking)
+const { Node } = require('prosemirror-model'); // eslint-disable-line no-unused-vars, (for typechecking)
+
+
+const shortnames = Object.keys(emojiByAllShortnames);
+
+/**
+ * @param {() => EditorView} getView
+ */
+function makeEmojiSuggestions(getView) {
```
anri-asaturov: Just curious about how PM works - does this get called on every character typed or only after trigger character?

src/ui/emoji/Picker.jsx
```diff
@@ -1,55 +1,30 @@
...
-    const catKeys = Object.keys(data);
-    for (let i = 0; i < catKeys.length; i++) {
-        data[catKeys[i]].forEach(item => {
-            shortnameMap[item.shortname] = item;
-            item.className = `emojione emojione-32-${getCategoryName(item)} _${item.unicode}`;
-        });
-    }
-}
+const recentList = computed(() => {
+    if (!User.current) return [];
```
anri-asaturov: I think it's safe to remove this check, User.current is never undefined after app route has changed to /app/*.
Not sure why I had a similar check before, maybe because I was pre-caching stuff even before login.

src/ui/chat/components/Message.jsx
```diff
@@ -170,4 +237,49 @@ class Message extends React.Component {
...
+/**
+ *
+ * @param {{ error : Error, info : { componentStack : string }}} errorData
+ * @param {*} msg TODO/TS: icebear message
+ */
+function renderError(errorData, msg) {
+    console.error('Error rendering the following message:');
+    console.dir(msg);
+    // If you change any tags or styles in here, make sure all text remains selectable
+    // (The CSS property "user-select" isn't inherited!)
```
anri-asaturov: Hmm, btw, would `.selectable * { user-select: text; }` help maybe? In a general sense of our app, not just in here.

src/ui/chat/components/Message.jsx
lostfictions: Yeah, agreed! We obviously can't localize the stack trace, payload dump, or what's being thrown, but I'll make sure everything else goes through the translator.

src/helpers/chat/prosemirror/chat-schema.jsx
```diff
@@ -0,0 +1,290 @@
...
+// (ie. https://developer.mozilla.org/en-US/docs/Web/API/Document/createTextNode) --
+// just be careful with attributes!
+const chatSchema = new Schema({
+    nodes: {
+        doc: { content: 'block+' },
+        paragraph: {
+            content: 'inline*',
+            group: 'block',
+            parseDOM: [{ tag: 'p' }],
+            toDOM() { return ['p', 0]; }
```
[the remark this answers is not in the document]
lostfictions: Yep, because PM is based on `contenteditable` (but maintains its own model) it needs a way to convert to the DOM in the input box.

The React renderer (used for displaying messages) can also use the intermediate representation returned by `toDOM()` to create the React version of the message if a `toReact()` method isn't present.

src/helpers/chat/prosemirror/chat-schema.jsx
```diff
@@ -0,0 +1,290 @@
...
+        },
+        blockquote: {
+            content: 'block+',
+            group: 'block',
+            defining: true,
+            parseDOM: [{ tag: 'blockquote' }],
+            toDOM() { return ['blockquote', 0]; }
```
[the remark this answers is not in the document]
lostfictions: Sure, we can just override everything with `toReact()` if it's genuinely a performance concern. I didn't do it because I hadn't profiled anything and having two duplicate functions expressing basically the same thing adds a maintenance burden (especially while the schema was in flux).

I'm also going to submit a RFC to ProseMirror to allow `toDOM` to be a field instead of a method -- that'll allow constant expressions like the one in this comment (ie, expressions that don't need extra calculations based on the node) to be fully resolved to React components ahead of time, which should make performance identical.

src/helpers/chat/prosemirror/chat-schema.jsx
```diff
@@ -0,0 +1,290 @@
...
+
+                        return false;
+                    }
+                }
+            ],
+            // HACK: we define both toDOM and toReact for emoji since React is
+            // stricter about nodes like <img> not having children -- but
+            // putting the emoji literal as a child of the img tag in the toDOM
+            // implementation lets us trivially get the emoji plaintext using
+            // the view element's .textContent.
```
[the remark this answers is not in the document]
lostfictions: I'll take another look at this -- I know that's what `alt` on img tags is supposed to do, but I think there was an issue.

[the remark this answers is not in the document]
lostfictions: Actually, just took a look and I remember -- for some reason unicode in the `alt` attribute doesn't show up? Will investigate.

src/helpers/chat/prosemirror/emoji-plugin.js
```diff
@@ -0,0 +1,99 @@
...
+                    // shortname for the combined sequence works, so we convert
+                    // to the image in two steps.
+                    const replacement = emojione.shortnameToImage(emojione.toShort(wrapped.text()));
+                    wrapped.replaceWith(replacement);
+                });
+                return $.html();
+            },
+            clipboardTextParser(text) {
+                const dom = document.createElement('div');
+                text.trim().split(/(?:\r\n?|\n)+/).forEach(block => {
```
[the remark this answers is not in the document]
lostfictions: This replicates ProseMirror's built-in behaviour on paste; I agree that it's easy to accidentally copy extra leading/trailing whitespace. For ourselves, we also decided internally not to allow this and there's a [TP US/task[(https://peerio.tpondemand.com/entity/7077) for it.

src/helpers/chat/prosemirror/emoji-plugin.js
```diff
@@ -0,0 +1,99 @@
...
+
+    return new Plugin({
+        props: {
+            // Using Cheerio here is maybe not the most optimal solution, but it
+            // works reliably and allows us to do a quick replacement on HTML
+            // without worrying that we're eg. invalidating the HTML by replacing
+            // inside an attribute value.
+            transformPastedHTML(html) {
+                // @ts-ignore bad typings
+                const $ = cheerio.load(html);
```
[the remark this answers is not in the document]
lostfictions: We could -- it would mean extra round trips to/from DOM (pasted HTML comes in as text, we convert to DOM and manipulate, convert back to text, ProseMirror converts it to DOM again) but I guess that's fine/not significantly less work from what Cheerio is doing.

More significantly, I think the implementation of that might be more complex than what's here, but I can take a look and see if it's simple. Agree that it would be nice to not have this dependency.

src/ui/chat/components/Message.jsx
lostfictions: In the props passed to each Message component? How does that help? Isn't this just on the prototype anyway...?

src/ui/chat/components/Message.jsx
lostfictions: Can you mark lifecycle methods as actions? I didn't want to mess with it.

src/ui/chat/components/Message.jsx
lostfictions: Yes, forgot to mention in the PR that's future work.

src/ui/chat/components/Message.jsx
lostfictions: Oh yeah, that seems like a good idea!

src/helpers/chat/prosemirror/chat-schema.jsx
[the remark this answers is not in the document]
anri-asaturov: Not only for the sake of performance, TBH (it does really need to be measured accurately first), but for clarity and simplicity. So, the current answer to "How do we render bold text" is "Easy, here - https://github.com/PeerioTechnologies/prosemirror-react-renderer/blob/master/src/index.ts" right? When it could have been 
```
toReact(){ return <strong>something</strong>}
````
Do I understand it correctly?

src/ui/chat/components/suggestions/EmojiSuggestions.jsx
lostfictions: Yeah, though I'd maybe beware of the enemy of all premature optimizations: actually causing a net loss in performance from blowing your cache by precalculating too much :P  The Emoji type is already pretty bloated with data we don't use, I think...

src/ui/chat/components/suggestions/EmojiSuggestions.jsx
lostfictions: Will fix

src/ui/chat/components/suggestions/EmojiSuggestions.jsx
lostfictions: This `makeEmojiSuggestions` is only getting called on construction in the MessageInput component; I also wrote the Suggestions component it constructs, so you can check what gets hooked into there. It's only the methods defined in the Plugin part of it that get callbacks (handleTextInput, etc.)

src/ui/emoji/Picker.jsx
lostfictions: Up to you -- I think it's not hurting maintainability/readability being defensive like this, and it shouldn't get recomputed too often so it's not really affecting performance.

src/ui/chat/components/Message.jsx
anri-asaturov: Hmm is it on prototype? I think babel puts it in constructor, will check, if yes then no problem ofc.

src/helpers/chat/prosemirror/chat-schema.jsx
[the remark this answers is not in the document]
lostfictions: Right, I guess if we're talking clarity and simplicity, the question comes down to "is it more confusing and complex to have the tag defined twice and needing to be updated in two places on any change, or can we trust the abstraction of the library code (which I wrote)?"

Feel free to answer that, but know that one of those answers might hurt me :)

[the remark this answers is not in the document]
lostfictions: Yes, `['blockquote', 0]` and `<blockquote>{children}</blockquote>` are expressing the same thing, and the latter can be derived from the former (which is it's doing now), and I think having them both if it's not necessary (especially if the `toReact` method gets moved out of the schema, which it will for cross-platform functionality) makes things trickier for anyone working with the code.

It's fine if you staunchly feel we should define both -- I don't disagree there may be valid reasons -- but deriving data like this instead of duplicating is pretty useful in the general case and I think should be preferred unless we really _have_ to duplicate something.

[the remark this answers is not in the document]
anri-asaturov: Given that it's literally 1 line of code that gets not duplicated, but reimplemented, I don't see much problem, maybe I don't understand everything yet. I'd like to summon @dchest's optinion and wait until we measure the performance.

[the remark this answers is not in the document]
lostfictions: Sure. I agree that it might just be bikeshedding over something relatively small at this point and we can just add `toReact` to everything -- I just don't feel great about adding gotchas to a codebase in general.

If we decide to add them, when migrating the schema to Icebear we could probably just keep the `toDOM` methods in Desktop alongside the `toReact` ones -- colocating them will at least make it more obvious that both need to change.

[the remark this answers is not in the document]
dchest: A-ha, that's why some contain `toReact` and some don't :) I missed that. I'd say, if there's not too much code to be saved by abstracting the duplication, and non-abstracted code is not too complicated, having both `toDOM` and `toReact` is better than having a magic transform from one to another.

src/ui/chat/components/Message.jsx
anri-asaturov: Yeah, you're right, it's in prototype and since it's not a bound action it doesn't initialize in every instance.

src/helpers/chat/prosemirror/chat-schema.jsx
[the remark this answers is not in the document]
lostfictions: Sounds good. Also, I'm realizing I never explained -- one reason I think it might be a gotcha is that if we migrate `chat-schema.jsx` to Icebear, we'd have to keep the `toReact()` methods in Desktop (since mobile will define their own `toReact`s). But `toDOM` is part of the schema specification, so it would end up in Icebear, and the duplicate definitions would be far from each other.

But I'll double-check if the schema will construct correctly without `toDOM` defined for all nodes. If so, than maybe what we can migrate to Icebear is just the spec object for constructing the schema, not the code that constructs the Schema instance itself. Then each consumer can merge the fields they need (`toDOM` and `toReact` on desktop, just `toReact` on mobile) into the spec and construct the schema from the merged object. That way we get to colocate the two methods on Desktop, and there's less irrelevant stuff in Icebear -- does that make sense?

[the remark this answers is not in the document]
dchest: > But I'll double-check if the schema will construct correctly without toDOM defined for all nodes. If so, than maybe what we can migrate to Icebear is just the spec object for constructing the schema, not the code that constructs the Schema instance itself. 

Agree, keeping just the schema without `toDOM` and `toReact` (if it's possible) in Icebear would be appropriate.

src/ui/chat/components/Message.jsx
anri-asaturov: Right, probably better not

src/ui/chat/components/suggestions/EmojiSuggestions.jsx
anri-asaturov: Yeah, nvm, this was over the top )

lostfictions:
@anri-asaturov okay, I've localized the error boundary. I think this is ready to merge -- I've noted down all the other fixes that should happen, but in the interests of keeping myself and Eren sane I think they should happen as separate PRs so we can more reasonably isolate what might have changed or would need extra testing. Sound good?

anri-asaturov:
@flohbot are we releasing desktop prosemirror before mobile? If not i think we should merge desktop and mobile at the same time?

lostfictions:
@anri-asaturov @flohdot since we have plaintext backwards/forwards compat, i think the plan is we're going to be releasing (initially read-only) rich text support on mobile after this is merged

lostfictions:
For reference, future work:
- Little fix: replace inline style in Emoji suggestion with class
- Assure forwards compatibility with future schema versions degrades gracefully

- Chores
  - Tag prosemirror-react-renderer version (and/or release on NPM) and make the entry in app/package.json point to it

- Cleanup:
  - Investigate using DOM APIs instead of Cheerio library for parsing pasted HTML
  - Investigate the need for  tags having an inner string when copy/pasting emoji in message input (maybe it's just not flagged as selectable?)

- Optimization/perf:
  - Cache message rendering results

- Icebear work:
  - Migrate the chat schema to Icebear (keeping the `toDOM` and `toReact` methods on the Desktop side)
  - Bring mention and URL handling in Icebear in line with the new message format behaviour

PR #108, opened 2017-11-08
fix: align inline file icon and make it part of clickable area

lostfictions:
Noticed this during the all hands and it was bugging me :sob: 

Before
before

After
after

------


suprko:
Haha looks good to me. This will probably change soon anyway, now that we have a fuller set of file-type icons :D

anri-asaturov:
OCD is good

PR #109, opened 2017-11-08
fix: inline files filenames should correctly truncate again

lostfictions:
Fix regression with long filenames:

screenshot_from_2017-11-08_13-14-59

------


src/ui/chat/components/InlineFiles.scss
```diff
@@ -89,7 +89,6 @@
 
   .inline-files {
     position: relative;
-    width: calc(100% - 100px); // reduce reflow from read receipts
```
lostfictions: Should no longer be necessary with fixed-size read receipt container (@anri-asaturov?)

PR #111, opened 2017-11-08
fix(sidebar): make chat purpose text selectable and fix ability to break layout when editing purpose

lostfictions:
Currently the sidebar purpose editor can be resized in a way that breaks layout:

screenshot from 2017-11-08 18-00-45

This PR fixes that, and also makes chat purpose text selectable (so stuff like URLs in the purpose can at least be copy-pasted even if they're not clickable yet)


src/ui/chat/components/sidebar/ChannelSideBar.jsx
```diff
@@ -81,8 +81,14 @@ class ChannelSideBar extends React.Component {
                         <div onClick={this.showChatPurposeEditor}>
                             {
                                 this.chatPurposeEditorVisible
-                                    ? <ChatNameEditor showLabel tabIndex="-1" purpose readOnly={!canIAdmin}
-                                        onBlur={this.hideChatPurposeEditor} multiline
+                                    ? <ChatNameEditor
```
anri-asaturov: I see you like long code files :-D

anri-asaturov:
Did you test this enough or shall i summon Eren?

src/ui/chat/components/sidebar/ChannelSideBar.jsx
lostfictions: you're right! in fact, i think we should also adopt object syntax like this
```
{ foo, bar,
  baz: 'value', barfly: true,
  gorf: this.bork, garple,
  dorg
}
```
:rofl:

lostfictions:
Maybe if he were up for taking a quick look that'd be a good idea? It shouldn't break anything but I didn't test super thoroughly

src/ui/chat/components/sidebar/ChannelSideBar.jsx
anri-asaturov: Absolutely hate it )

anri-asaturov:
@ebilg the scene is yours

lostfictions:
Actually, I just took another quick look at this and tried to break it -- I don't think Eren necessarily needs to look at it since there's no real behaviour changes other than the resize grippy being gone. Should be ok to merge.

PR #120, opened 2017-11-13
fix: make whole message send error clickable, not just the icon

lostfictions:

## PR author

|    📝   |
| ------- |
| ....... |
| ....... | 
| ....... |
| **Breaking changes:** | 
| ....... |
| **Additional notes:** | 
| ....... |

## Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

## Repository owner


src/ui/chat/components/Message.jsx
```diff
@@ -43,12 +43,23 @@ const UrlPreviewConsent = require('./UrlPreviewConsent');
 @observer
 class Message extends React.Component {
     @observable.shallow errorData = null;
+    @observable errorMenuVisible = false;
```
lostfictions: Needed to add state field because RT `<Menu>` doesn't track its own state, unlike `<IconMenu>`

src/ui/chat/components/Message.jsx
```diff
@@ -203,18 +214,23 @@ class Message extends React.Component {
...
                                 </div>
-                                <div className="send-error-message">{t('error_messageSendFail')}</div>
+                                <Menu
+                                    position="auto"
+                                    ripple
+                                    active={this.errorMenuVisible}
+                                    onHide={this.hideErrorMenu}>
+                                    <MenuItem
+                                        caption={t('button_retry')}
+                                        onClick={() => m.send()} />
```
anri-asaturov: hmm, probably better to have an arrow function here then bound action in every message, right?

anri-asaturov:
Do you need Eren to test it?

src/ui/chat/components/Message.jsx
lostfictions: yeah true

lostfictions:
should be good -- the only slightly weird thing i noticed is that the menu was positioned way off to the side, but i just fixed it.

PR #122, opened 2017-11-13
fix: make ascii sequences like ;) get replaced by their emoji versions again

lostfictions:
Fixes https://peerio.tpondemand.com/entity/7448.

~~Relatively simple, but I'm thinking `src/helpers/chat/emoji.js` and `scripts/build-emoji.js` should probably be consolidated so that as much work is done as possible ahead of time (and perhaps so that unused data in the processed JSON file is removed). Using `console.time()`/`console.timeEnd()`, it seems like this PR adds over a second of startup time, which really isn't great.~~
This PR now includes a refactor to `scripts/build-emoji.js` to do as preprocessing as possible.

I also think this replacement action should be behind a user preference, so that users who just want to type `:)` don't always have to undo the replacement.

## PR author

|    📝   |
| ------- |
| https://peerio.tpondemand.com/entity/7448 |
| ....... | 
| - Emoji sequences like `;)` should convert after typing them and hitting space. (See issue on TP for list of sequences.) |
| - The conversion shouldn't happen under any circumstances that might be accidental (writing closing parens, etc.) |
| **Additional notes:** | 
| ....... |

## Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

## Repository owner


anri-asaturov:
The setting is easy to add (look at `UIStore.prefs`) as soon as you clear up with Paul where exactly should the toggle be.

anri-asaturov:
btw, emojione does this detection/conversion any reason not to use it?

lostfictions:
Should it be a local setting or a synced one?

anri-asaturov:
Question for @paulrdukes

lostfictions:
@anri-asaturov the emojione conversion is inline, only converts to shortcodes/images; we want an input rule that applies after hitting space (unlike the shortcode behaviour), is undoable, and converts directly to an emoji node.

lostfictions:
@anri-asaturov refactored the build script -- let me know if that second commit seems ok to you. otherwise this should be good to merge, i'll make a note to ask Paul about the pref

lostfictions:
oops, belatedly realizing this won't work for the case of hitting eg. `:)` and then enter immediately -- will fix tomorrow.

lostfictions:
TP bug/info in https://peerio.tpondemand.com/entity/7448

PR #125, opened 2017-11-15
fix: put chatName variable in drop target dialog

suprko:

## PR author

|    📝   |
| ------- |
| https://peerio.tpondemand.com/entity/7460 |
| n/a | 
|tested and approved|
| **Additional notes:** | 
|n/a|

## Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

## Repository owner


src/ui/shared-components/DropTarget.jsx
```diff
@@ -56,7 +56,13 @@ class DropTarget extends React.Component {
                     onEscKeyDown={this.cancelUpload}
                     onOverlayClick={this.cancelUpload}
                     title={t('title_uploadAndShare')}>
-                    <p>{t('title_fileWillBeShared')}</p>
+                    <p>{t('title_fileWillBeShared',
+                        {
+                            chatName: chatStore.activeChat.isChannel
```
anri-asaturov: activeChat can be null

suprko: I just tested, if activeChat is null then this dialog never appears.

(Though there is a weird behaviour to check out: if activeChat is null, you can still upload files, and it just goes straight to your Files, not shared with any chat. I don't know, maybe that's OK)

lostfictions: Yeah, looks like it's [this](https://github.com/PeerioTechnologies/peerio-desktop/blob/181c936ae883e4a48977f8a1728ce66b4bea3b64/src/ui/shared-components/DropTarget.jsx#L27) doing that -- but to be honest to me that's really surprising control flow. Different routes should probably define their own drop targets rather than having a single global drop target that checks what route it's in.

anri-asaturov: We have pretty much 1 target except default, so doesn't seem important enough to do the override logic (you'll need to hide global one under specific routes that have their own anyway right).

lostfictions: Not sure I agree. I think we have two pretty different behaviours now (upload and share to chat, upload into your files) and the behaviour is only going to diverge more with folders, right?

The more divergent the behaviour can be, the less it makes sense to me to have a global overlay -- eg. in the contacts view, what should uploading do? Can you drag a file over a contact to upload and share with them? What would a user reasonably expect dropping a file in that view to do?

anri-asaturov: No, don't take me wrong i'm not against it all, just explaining the current state of it. Folders make a good case for changing it.

lostfictions: Yeah, I don't think it's something that needs to change right now! I get why it's this way.

And it's definitely not blocking the current PR -- this all LGTM.

PR #130, opened 2017-11-16
fix: resolve sidebar menu item and display issues

lostfictions:
Three small fixes:

- User actions menu is no longer clipped
- Long usernames and file names will correctly be truncated with ellipses
- Chat title text is selectable, like chat purpose text

You can look at each of the three commits atomically -- they shouldn't affect each other.

Fixes [TP#5603](https://peerio.tpondemand.com/entity/5603), [TP#7556](https://peerio.tpondemand.com/entity/7556)

----

### Before:

usernames-before

screenshot from 2017-11-15 17-05-31

users-before

invited-before

### After:

screenshot from 2017-11-16 17-31-57

screenshot from 2017-11-16 17-40-23

screenshot from 2017-11-16 17-04-03

screenshot from 2017-11-16 17-32-53

----

## Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

## Repository owner


lostfictions:
@ebilg possible regressions: @suprko would know more, but it seems like the original css that was causing the menu clipping was intended to prevent some kind of overflow on the x-axis, but i can't seem to make that happen no matter what i try.

anri-asaturov:
Does the member section still scroll when overflowed?

suprko:
K I'm pretty sure I know what happened. The whole member list was able to be scrolled horizontally, so I added an overflow-x hidden, but that caused individual member entries to be scrollable horizontally somehow. After digging around a bit, I solved that by putting an explicit width on the member list, but I didn't remove the overflow rule.

Anyway. Looks like everything works and nothing has regressed.

PR #131, opened 2017-11-17
Add identity verification warning

alexandra-marin:

## PR author

|    📝   |
| ------- |
| https://peerio.tpondemand.com/RestUI/Board.aspx#page=task/7407&appConfig=eyJhY2lkIjoiMTA1MTA5MDU0OEY2QTUyQjlFM0JCODkwRjYwQUVGMEIifQ== |
| no | 
| 1. Create DM - warning should be displayed on top 2. Create room - warning should be displayed on top and when someone joins |
| **Additional notes:** | 
| no |

## Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

## Repository owner


src/ui/chat/components/IdentityVerificationNotice.jsx
```diff
@@ -0,0 +1,16 @@
+const React = require('react');
+const T = require('~/ui/shared-components/T');
+const { FontIcon } = require('~/react-toolbox');
+
+class IdentityVerificationNotice extends React.Component {
+    render() {
+        return (
+            <div className={`notice ${this.props.extraMargin && 'leftMargin'}`}>
```
anri-asaturov: i'm not sure u want `undefined` or `false` class to apply here.

src/ui/chat/components/Message.jsx
```diff
@@ -102,6 +103,14 @@ class Message extends React.Component {
...
         if (!m.systemData) return null;
+        if (m.systemData.action === 'join') {
+            return (
+                <div className="system-message selectable">
+                    <p>{systemMessages.getSystemMessageText(m)}</p>
+                    <IdentityVerificationNotice extraMargin="true" />
+                </div>
+            );
+        }
         return <p className="system-message selectable">{systemMessages.getSystemMessageText(m)}</p>;
```
anri-asaturov: This markup doesn't look like a good idea, it's not consistent. @suprko please guide us.

suprko: @anri-asaturov are you referring to the fact that `div` or `p` are assigned the same classNames, depending on the bool?

If so my solution would be to put the conditional inside the `return` probably, something like this
```
    renderSystemData(m) {
        // !! SECURITY: sanitize if you move this to something that renders dangerouslySetInnerHTML
        if (!m.systemData) return null;
        return (
            <div className="system-message selectable">
                <p>{systemMessages.getSystemMessageText(m)}</p>
                {m.systemData.action === 'join' &&
                    <IdentityVerificationNotice extraMargin="true" />
                }
            </div>
        );
    }
```

anri-asaturov: Yes, your solution

src/ui/chat/components/IdentityVerificationNotice.jsx
suprko: @alexandra-marin take a look at the `classnames` component, it's really useful for conditional classes like this.

e.g. AppNavButton has this:
`div className={css('menu-item', { active: this.props.active })}`

src/ui/chat/components/IdentityVerificationNotice.scss
```diff
@@ -0,0 +1,22 @@
...
+
+    .icon {
+        margin-right: $margin-small;
+    }
+
+    a {
+        color: $text-dark-secondary !important;
+        text-decoration: underline;
+    }
+}
```
lostfictions: These styles should probably be wrapped in a more specific class name, right @suprko? Currently all of our CSS on Desktop is in the global scope, despite the modularized files :(

suprko: @alexandra-marin Anri will make you do kebab case for everything :D so for example `identity-verification-notice` instead of `identityVerificationNotice`

In terms of the SCSS scoping that @lostfictions mentioned above, what we've been doing on other files is basically wrap the entire SCSS document in a rule that is specific to that component. So in this case, we'd begin the file with `.identityVerificationNotice {`, and put all of our rules inside there, so that we know they won't affect other components. So after that, it's OK if you use more generic classes like `.icon` or `.extra-margin` because the scoping will limit their effect.

suprko: ^ just to clarify, kebab case for CSS classnames. Still camel case for JS variables :)

lostfictions: It's not just Anri complaining about style for no reason! Standard style for CSS is kebab case because selectors are case-insensitive, so `coolClassName` is semantically the same as `coolclassname` -- potentially a source of bugs. (And on the other hand, the reason we don't use kebab case in JS is because it causes syntax errors :) )

anri-asaturov: I personally have no preference for naming style as long as it's consistent across the repo.

alexandra-marin: @lostfictions thanks for the explanation, helpful to know why this rule exists :)

src/ui/chat/components/IdentityVerificationNotice.scss
```diff
@@ -0,0 +1,25 @@
...
+    .extra-margin {
+        margin: 0 0 0 $margin-default;    
+    }
+
+    .notice-icon {
+        margin-right: $margin-small;
+    }
+
+    a {
+        color: $text-dark-secondary !important;
```
lostfictions: Sorry, still needs a fix or two :(

You should never have to use `!important` in our own CSS. I'd recommend thinking of it as something like the `debugger;` statement in JS -- sometimes useful when developing, but it should never ship in production code or be merged into a main branch.

If you're not sure how specificity in CSS works (understandable! It's definitely not always intuitive) I'd recommend reading [this](https://css-tricks.com/specifics-on-css-specificity/) for a good primer. It's not always easy to remember the rules off the top of your head even if you understand the principle -- I also find http://cssspecificity.com/ is a very useful quick reference.

All of this to say, now that this element block is wrapped in a class block, this rule has enough specificity that it should work without `!important` -- does that make sense?

lostfictions:
Also, this totally isn't your fault at all, but the CI linter hook wasn't parsing all files correctly -- I just fixed in on dev -- and so it wasn't catching the wrong indentation on the CSS file. (We use two spaces in CSS but four spaces in JS, but I'm not sure why.) Anyway, please merge in dev and double-check that everything still passes the validation step again!

src/ui/chat/components/Message.jsx
```diff
@@ -102,7 +103,15 @@ class Message extends React.Component {
     renderSystemData(m) {
         // !! SECURITY: sanitize if you move this to something that renders dangerouslySetInnerHTML
         if (!m.systemData) return null;
-        return <p className="system-message selectable">{systemMessages.getSystemMessageText(m)}</p>;
+        return (
+            <div className="system-message selectable">
+                <p>{systemMessages.getSystemMessageText(m)}</p>
+                {
+                    m.systemData.action === 'join' &&
+                    <IdentityVerificationNotice extraMargin="true" />
```
lostfictions: BTW, in JSX `<IdentityVerificationNotice extraMargin="true" />` is equivalent to just `<IdentityVerificationNotice extraMargin />`

lostfictions: Is there a reason the `extraMargin` needs to be configurable/passed in as a prop to the component?

alexandra-marin: depending on where i use the component, i need the extra spacing. it didn't make sense to have it in the parent component

lostfictions: Oops -- nevermind. Just saw the other place the component is rendered without it :flushed:

src/ui/chat/components/IdentityVerificationNotice.scss
alexandra-marin: Wasn't happy with that either, but didn't find any other solution. 
I get the element from a T (span > a), which means I can't add inline style or an id. Even if I do .identity-verification-notice > .notice-container > a it still gets overridden...

suprko: This is more of a design problem but we have a weird inconsistency with text links sometimes being blue and sometimes being grey. Perhaps worth bringing it up with Paul.

So... it seems a little silly but you could apply a className to `<T>` itself (e.g. `text-content`) and then add a rule for `.text-content > a`

lostfictions: @alexandra-marin is right -- there's a higher-specificity override because of the message view rules:

screenshot from 2017-11-20 13-41-17

This is another big reason why we really, really need scoped CSS. The more styles we write the more painful it's going to get -- splitting stuff out to files like this really doesn't help scale CSS.

alexandra-marin: @suprko Tried that too, gets overridden

lostfictions: Yeah, it's really just a case of another rule having three class selectors and an element selector. This extremely sucks.

suprko: Working for me like this:
`<T k="title_verifyUserIdentity" className="text-content" />`

```
.notice-container {
    (bunch things...)

    .text-content > a {
      color: $text-dark-secondary;
    }
  }
```

alexandra-marin: Ah, yes. I didn't nest it as much. Fixed!

suprko: Glad it's fixed! For the record I absolutely agree with @lostfictions that our CSS situation is a rat's nest but of course that is far outside the scope of this PR :D

lostfictions: Thanks for the change! Definitely this is a problem with our toolchain/setup and I think !important is understandable as an approach in this situation (it's easy to imagine a scenario where the elements are nested slightly less or couldn't have a classname assigned to them and there wouldn't be _any_ other solution besides !important).

anri-asaturov:
Waiting for @ebilg 's review

PR #133, opened 2017-11-20
feat: sort folders by name instead of date

seavan:


anri-asaturov:
Small change, no testing required

PR #144, opened 2017-11-23
feat(security): block inline and external scripts and styles via Content-Security-Policy

dchest:
No external or inline scripts or styles will be allowed to prevent many classes of XSS (we can later relax this for inline styles, if we'll use them, but better not — instead use a proper inline->classes style compiler).

It also disallows the use of `eval()`.

...

...

**Everything** needs to be tested.

#### Additional notes
... 

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


anri-asaturov:
Are you sure this doesn't break anything? :-D

dchest:
Not completely.  That's why we should test everything 🌝  But I think we don't have any inline scripts or styles.

anri-asaturov:
We do have inline styles in react. Why do we want to block inline styles tho?

dchest:
For best results, yes, since there are some possibilities for XSS via CSS. If we have inline styles, then we should turn then on until we switch to some kind of React CSS module that generates classes.

So, change `style-src 'self'` to `style-src 'self' 'unsafe-inline'`.

lostfictions:
Haven't tested this yet, but it seems like a good idea!

I've never used CSP in the context of an Electron app, but presumably it applies to everything called from the renderer process?

One thing I noticed is that devtools get disabled by the CSP -- not sure why, maybe the way the devtools installer injects them?

screenshot from 2017-11-23 11-49-09
screenshot from 2017-11-23 11-59-48

Seems they can be whitelisted in the CSP, but Chrome doesn't recognize the protocol as a wildcard (so `script-src 'self' chrome-extension://react-perf;` works, but `script-src 'self' 'chrome-extension:'` doesn't)

lostfictions:
Oops, this got merged while I was writing a comment about how it breaks stuff :joy:

dchest:
Ah, right; we'll have to whitelist all these tools, or maybe just for development version.

dchest:
what about `chrome-extension://*` ?

lostfictions:
Tried that too, didn't seem to work (though I quoted it, maybe that was the problem)

lostfictions:
Yeah, seems to work unquoted actually!

dchest:
Probably: [single quotes shouldn't be used](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src)

lostfictions:
ugh, can't add commits to this branch because of the squashed merge. opening a new branch/pr...

lostfictions:
The other thing I wanted to ask is whether it'd be a good idea to further restrict some stuff as long as we're setting this now? `object-src`, `frame-src`, `worker-src`?

dchest:
Yes, indeed. Small steps :)

I checked and see that there are a few inline styles that we can't now replace with classes (e.g. `backgroundImage` for avatar), so we'll have to enable `'unsafe-inline'` for them.

lostfictions:
@anri-asaturov, better if I make a new PR rather than you reverting this merge?

dchest:
Actually, yes, let add everything now (except for `image-src`) and then we'll narrow down policies if something breaks.

PR #145, opened 2017-11-23
fix: attempt to fix stuff broken by csp branch merge + add more rules

lostfictions:
we could also try `default-src: 'self'` and whitelist images, media, sockets, scripts, and styles, but i'm not sure that's better :)


src/index.html
```diff
@@ -2,7 +2,10 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <meta http-equiv="Content-Security-Policy" content="script-src 'self'; style-src 'self'">
+    <meta
+        http-equiv="Content-Security-Policy"
+        content="script-src 'self' chrome-extension:; style-src 'self' 'unsafe-eval' 'unsafe-inline'; object-src 'self'; frame-src 'self'; worker-src 'self';"
```
dchest: Is 'unsafe-eval' needed for styles?

lostfictions: Going by MDN:
> The 'unsafe-eval' source expression controls several style methods that create style declarations from strings. If 'unsafe-eval' isn't specified with the style-src directive, the following methods are blocked and won't have any effect:
>
>    CSSStyleSheet.insertRule()
>    CSSGroupingRule.insertRule()
>    CSSStyleDeclaration.cssText

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src

lostfictions: Not sure if any library code we have uses any of those right now, but I figured we should take the option that will potentially unbreak the most stuff now since the previous PR is already merged :)

dchest: 👍

anri-asaturov: I think we should make it as strict as possible and monitor next week, people will complain and we'll fix. It is really hard to anticipated all the possible issues and test them in one go too.

PR #147, opened 2017-11-23
chore: React console errors on prod only

suprko:

https://peerio.tpondemand.com/entity/7506
n/a
Quick test by me, not yet by @ebilg 

Some of these changes are pretty broad, for example pretty much the entire app is wrapped in a new `<div>` in both App.jsx and Root.jsx, so there is a small potential for layout to break, well, pretty much anywhere. (It will be obvious if so). Because I needed to look through every view to check for console errors anyway, I should have caught these cases, but I may have missed something.

In order to test this you need to go into `package.json` and add `NODE_ENV=production` to the "start" script (first line below "MAIN COMMANDS"), so the full line reads `"start": "NODE_ENV=production npm run compile && npm-run-all -l -p code:watch icebear:watch style:watch devstyle:watch electron:run"`

#### Additional notes
n/a

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


lostfictions:
Gotta say, I'm legit kinda baffled by all of these changes. None of these should be necessary -- sibling div elements shouldn't need keys, for example. Does this really squelch the errors? Have we figured out where exactly they're coming from and why they're only happening in production?

I kinda don't think this PR should be merged until we at least figure out the answer to these questions, if only so we can be sure that they won't continue to happen again by accident -- assuming it's our fault.

anri-asaturov:
@lostfictions yeah i think we should just ignore unexplainable key warnings until we move away from react-toolbox, it seems to cause most of the issues.

suprko:
@lostfictions they do get rid of the errors but I was basically shooting in the dark because I don't fundamentally actually understand why the errors are happening, and specifically why in `production` but not `development`.

@anri-asaturov can you tell us more about what these errors are and where they're coming from? I don't see any indication that react-toolbox components are actually the source but I don't understand enough about the errors themselves to know... anything

anri-asaturov:
Lucas, normally, react detects when you pass an array of similar children components and warns you that you will benefit from assigning unique keys to them so it can optimize render (reuse nodes).

But this case... is plain weird, and I suspect it's react-toolbox/themer which haven't been migrated to react 16 yet creates this effect. It's especially weird that it happens in prod env, it totally shouldn't. So I suggest to remove RTB first and then review the situation. I don't think it's worth the time waste otherwise.

PR #149, opened 2017-11-24
feat: retry loading images after connection restored

alexandra-marin:

https://peerio.tpondemand.com/RestUI/Board.aspx#page=bug/7463&appConfig=eyJhY2lkIjoiMTA1MTA5MDU0OEY2QTUyQjlFM0JCODkwRjYwQUVGMEIifQ==
no 
1. open app, close internet connection before app fully loaded
2. images should be replaced by 'broken link' symbol
3. open internet connection, images should reload automatically
#### Additional notes
Applied the changes in other places images are loaded from urls, not just avatars.

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


src/ui/chat/components/ReloadingImage.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+@observer
+class ReloadingImage extends React.Component {
+    assignImageRef = (ref) => { this.image = ref; };
+
+    retryDownload = () => { this.image.src = this.props.url; };
+
+    onError = (err) => {
+        if (err) {
+            when(() => uiStore.isOnline, this.retryDownload);
```
anri-asaturov: Please, store the disposer of this reaction and call it in `componentWillUnmount`. Also, just in case, don't add additional reaction if onError was called while reaction already exists.

src/ui/chat/components/ReloadingImage.jsx
```diff
@@ -0,0 +1,32 @@
...
+const { observer } = require('mobx-react');
+
+@observer
+class ReloadingImage extends React.Component {
+    assignImageRef = (ref) => { this.image = ref; };
+
+    retryDownload = () => { this.image.src = this.props.url; };
+
+    onError = (err) => {
+        if (err) {
```
anri-asaturov: What's the motivation for this check? Can onError get called without error object and we consider that not an error?

src/ui/contact/components/ContactProfile.jsx
```diff
@@ -63,8 +61,7 @@ class ContactProfile extends React.Component {
                     </div>
                     : null}
                 <div className="contact-avatar-info">
-                    {/* TODO: use our Avatar component instead, and strip it of mouse events */}
```
anri-asaturov: @suprko 👍 👍 👍

src/ui/settings/components/SecuritySettings.jsx
```diff
@@ -195,7 +196,7 @@ class SecuritySettings extends React.Component {
                                         ? <div>
                                             <T k="title_scanQRCode" tag="div" />
                                             <br />
-                                            <img alt={this.twoFASecret} src={this.twoFAQRCode} />
+                                            <ReloadingImage alt={this.twoFASecret} url={this.twoFAQRCode} />
```
anri-asaturov: This is not an external image, it uses data-url generated in client always.

alexandra-marin: sorry, looked like an url

src/ui/chat/components/ReloadingImage.jsx
alexandra-marin: it's extra, i'll remove it

src/ui/settings/components/SecuritySettings.jsx
anri-asaturov: no need to be sorry, it's a normal process :-P

src/stores/ui-store.js
```diff
@@ -43,6 +43,10 @@ class UIStore {
         prereleaseUpdatesEnabled: false
     }
 
+    @observable isOnline = false;
```
anri-asaturov: Oh. the default should be `navigator.onLine`, no?

src/stores/ui-store.js
```diff
@@ -43,6 +43,12 @@ class UIStore {
         prereleaseUpdatesEnabled: false
     }
 
+    @observable isOnline = navigator.onLine;
+
+    @action.bound changeOnlineStatus() {
+        this.isOnline = navigator.onLine;
```
lostfictions: btw, `navigator.onLine` only detects whether you're connected to a network or not, not whether you're actually online, so it'll fail in a lot of contexts. might be worth revisiting since there's lots of parts of the app that would benefit from improved connectivity detection

anri-asaturov: Yeah, we have a US somewhere for improved connectivity detection. That includes aggressive ping on waking up.

lostfictions: :+1:

anri-asaturov: That said i'll probably add our socket reconnect event to this.

src/ui/chat/components/ReloadingImage.jsx
```diff
@@ -0,0 +1,40 @@
...
+class ReloadingImage extends React.Component {
+    disposeReloadListener = null;
+
+    assignImageRef = (ref) => { this.image = ref; };
+
+    retryDownload = () => { this.image.src = this.props.url; };
+
+    onError = () => {
+        if (!this.disposeReloadListener) {
+            this.disposeReloadListener = when(() => uiStore.isOnline, this.retryDownload);
```
anri-asaturov: Ups, this actually will create infinite retry loop on non-image urls or http 404. :-D i'll fix it

alexandra-marin: why? doesn't the check for an already active listener take care of that?

anri-asaturov: Yeah, you are right, the issue is that it won't retry more than once, the check takes care of the loop, we could fix that with clearing the disposer variable tho.

anri-asaturov: Another problem in here is the current state on the moment of render, first of all you need to use socket.connected as it is more reliable, and then if onError runs after connection is restored - you won't retry at all. Weird timing issues.

PR #151, opened 2017-11-24
feat: jitsi videocalls

gembarrett:

### PR author
Gem

button disabled: https://peerio.tpondemand.com/entity/7592
implement UI: https://peerio.tpondemand.com/entity/6782

https://github.com/PeerioTechnologies/peerio-icebear/pull/60
https://github.com/PeerioTechnologies/peerio-mobile/pull/65

n/a

#### Additional notes
n/a 

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


src/ui/chat/ChatView.jsx
```diff
@@ -120,6 +120,14 @@ class ChatView extends React.Component {
         if (ref) ref.nameInput.focus();
     };
 
+    postJitsiLink = () => {
+        let randomString = crypto.cryptoUtil.getRandomUserSpecificIdB64(crypto.cryptoUtil.getRandomNumber(42, 1337));
+        randomString = randomString.replace(/(.{36}$)/, '').replace(/(\/)/, '+');
+        const jitsiLink = `https://meet.jit.si/${randomString}`;
+        this.selfNewMessageCounter++;
+        chatStore.activeChat && chatStore.activeChat.createVideoCall(jitsiLink);
+    };
```
anri-asaturov: Ok, 
1. this function, both desktop and mobile. There's some code duplication, please replace lines 124-126 with 
```
const jitsiLink = chatStore.generateJitsiUrl();
```
2. We have more elegant way of producing url-safe ids and if you want to use user-specific ids, pass current user's username instead of random number.

Icebear in `chat-store.js` add this
```
generateJitsiUrl(){
   const id = crypto.cryptoUtil.getRandomUserSpecificHex(User.current.username);
   return `https://meet.jit.si/${id}`;
}
```
If someone finds this id too long (although it's not much longer than b64 version) - tell me, i'll make a function to generate shorter id.

src/ui/chat/components/Message.jsx
```diff
@@ -103,6 +103,20 @@ class Message extends React.Component {
     renderSystemData(m) {
         // !! SECURITY: sanitize if you move this to something that renders dangerouslySetInnerHTML
         if (!m.systemData) return null;
+        if (m.systemData.link) {
+            const videoCall = m.systemData.link;
+            const videoCallShort = videoCall.replace(/(https:\/\/)/, '');
```
anri-asaturov: Why regexp tho? `videoCall.replace('https://', '');` does the same and easier to read.

anri-asaturov: On the other note, maybe we should just store call ID, what if we decide to host our own jitsi server or switch to some other videocall service. @flohdot would we rather show constant link or allow ourselves to replace domain for older videocalls?

src/ui/chat/components/Message.jsx
```diff
@@ -103,6 +103,20 @@ class Message extends React.Component {
...
+            const videoCallMsg = systemMessages.getSystemMessageText(m);
+            return (
+                <div>
+                    <p className="video-system-message">{videoCallMsg}</p>
+                    <p>
+                        <FontIcon value="videocam" className="video-icon" />
+                        <Link href={videoCall}>{videoCallShort}</Link>
```
anri-asaturov: I suspect it might not be selectable, apply 'selectable' class if i'm right.

src/ui/chat/components/Message.jsx
flohdot: I think we should stick with constant links. An old link will be useless (possibly to the point of confusion) if we do move.

anri-asaturov: ok, yeah, i guess it's not really a problem if your old video links won't work anymore. Just replace regex with string then @gembarrett please

src/ui/chat/ChatView.jsx
flohdot: eh my bad, I suggested the base64 version, forgot we had hex which would indeed be better

src/ui/chat/components/Message.jsx
gembarrett: The link works as expected, icon does not need to be selectable.

flohdot: the text is selectable. it's fine.

anri-asaturov:
@gembarrett reminder that PR titles should be in commit format, because we squash on merge

src/ui/chat/components/MessageList.scss
```diff
@@ -185,6 +185,45 @@
...
+
+      s {
+        @extend %selectable;
+      }
+
+      b,
+      strong {
+        @extend %selectable;
+        font-weight: $font-weight-semibold;
+      }
```
anri-asaturov: @suprko @gembarrett  Somehow all this styles don't strike me as required for jitsi? Could you verify?

anri-asaturov: @suprko could you verify this styles please, they're not scoped AFAIU

suprko: These rules look like they're copied from `_ChatComponentsCommon`, @gembarrett were they not being applied correctly from there?

gembarrett: I don't know, I didn't add those lines of code - I expect they were brought in by a merge. Looking at Git blame, the lines were added by @lostfictions so they should be asked for the rationale.

suprko: Arg, I suspect we have another git branch-merge-order situation. These were added by @lostfictions but removed in this commit 2 weeks ago and moved to `_ChatComponentsCommon`: https://github.com/PeerioTechnologies/peerio-desktop/commit/a625135da79da2f8615b8f493531cc4181ea1544

But it looks like they're showing up here again as new commits. @seavan we may need your git wizard eyes. Or alternatively we just remove them and move forward with the merge, since it seems to be isolated to one part of one file.

anri-asaturov: Could you just fix it in this branch please, Lucas?

anri-asaturov: @gembarrett Did you have any merge conflicts on this branch before? Maybe that's when deleted stuff stayed.

src/ui/chat/components/Message.jsx
```diff
@@ -103,6 +103,20 @@ class Message extends React.Component {
     renderSystemData(m) {
         // !! SECURITY: sanitize if you move this to something that renders dangerouslySetInnerHTML
         if (!m.systemData) return null;
+        if (m.systemData.link) {
```
anri-asaturov: Missed this on review (already fixed in dev)
If any other action type will have data property named `link`, this will fail.
Correct check would be `m.systemData.action === 'videoCall'`

PR #152, opened 2017-11-26
feat: network management improvements

anri-asaturov:
This PR solves several connection management issues.

1. Speedy recovery of the connection after OS resume.
If OS was asleep for long enough, on resume socket connection won't be working immediately, there will be a 15-30 second lag. So we make a request with short timeout right after resume and force reconnection earlier.

2. html `img` elements that failed to load due to disconnection will be forced to reload on reconnect.

3. Sometimes(quite often) chromium already knows that network is disconnected (navigator.onLine), but websocket is gonna be open for a while, resulting in a noticeable lag on reconnect. Now we force socket close when chromium reports offline event.


anri-asaturov:
@ebilg I know this is hard to test, and I did it myself, but you can at least find related TP entities and close them since I made all this changes at once while fixing Alexandra's PR.

src/helpers/image-retry.js
```diff
@@ -0,0 +1,28 @@
...
+    const elements = document.getElementsByTagName('img');
+    for (let i = 0; i < elements.length; i++) {
+        const img = elements[i];
+        // complete - means it's currently not loading
+        // naturalHeight - will be 0 if loading failed
+        if (!img || !img.complete || img.naturalHeight > 0) continue;
+        // not interested in local URIs
+        if (img.src.startsWith('file://')) continue;
+        // magically making image reload
+        img.src = img.src;
```
flohdot: Witchcraft. (Why does this work?!)

anri-asaturov: It's a setter, i think.

dchest: In some cases, like avatars we have images loaded as CSS `background`, so this won't reload them and I don't think we can? (Anyway, it would be good to get rid of remote background images for the sake of CSP if we can).

anri-asaturov: Yeah avatars are already changed to use `img`

dchest: Oh, right, I should have read the rest of the PR :)

anri-asaturov:
@flohdot well, many reasons, you can check the original PR (https://github.com/PeerioTechnologies/peerio-desktop/pull/149) 
It has a bug, which wasn't too easy to fix, but I don't like the whole approach(which was my idea originally tho) anyway.

- onError event doesn't give you any details on why did it fail, so it's not more useful
- component approach is less performant, even if not very noticeably 
- you have to remember to wrap your images with the component
- new approach covers non-react nodes, don't think we have them currently but maybe we will someday

src/helpers/network.js
```diff
@@ -0,0 +1,19 @@
...
+// Websocket fails to properly detect connection loss and this causes noticeable lags
+// when user experiences a brief disconnection or network switch.
+// Sometimes websocket connection won't break even if you disable and then enable wifi in a few seconds.
+// It will keep same connection but the actual data in it will start flowing with a big lag of 20-40 seconds.
+// So we decide to forcibly close the connection when chromium reports offline event, and then reopen it when
+// online event is fired.
+reaction(() => appState.isOnline, isOnline => {
+    if (isOnline && !socket.connected) {
+        socket.open();
+    } else if (!isOnline && socket.connected) {
```
lostfictions: Do we need the extra `&& !socket.connected` / `&& socket.connected` here? Closing and opening the socket isn't idempotent?

src/helpers/power.js
```diff
@@ -4,11 +4,21 @@ const { socket } = require('~/icebear');
...
 electron.powerMonitor.on('resume', () => {
     console.log('The system is going to resume');
     if (!socket.authenticated) return;
-    socket.send('/auth/push/disable');
+    socket.send('/auth/push/disable')
+        .timeout(3000) // short timeout here will let us avoid long lags in connection that happen after sleep
+        .catch(err => {
+            console.error(err);
+            console.log('Connection seems to be broken after OS resume, forcibly reconnecting to avoid lag.');
+            socket.reset();
```
lostfictions: After the socket is reset, is push then disabled? Couldn't find any code in icebear/desktop/this PR related to this.

src/ui/chat/components/InlineFiles.jsx
```diff
@@ -131,7 +130,7 @@ class InlineFile extends React.Component {
             <Dialog active={this.imagePopupVisible} type="large" ref={this.onPopupRef}
                 onOverlayClick={this.hideImagePopup} onEscKeyDown={this.hideImagePopup}
                 className="image-popup">
-                <ReloadingImage url={this.currentImageSrc} />
+                <img src={this.currentImageSrc} alt="" />
```
lostfictions: Why the empty `alt` attributes?

src/helpers/image-retry.js
```diff
@@ -0,0 +1,28 @@
...
+ * 3. It forces reload of failed images.
+ */
+
+const { socket } = require('~/icebear');
+const { reaction } = require('mobx');
+
+reaction(() => socket.connected, connected => {
+    if (!connected) return;
+    // caution: this is not an array, no forEach is available
+    const elements = document.getElementsByTagName('img');
```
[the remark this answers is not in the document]
lostfictions: (If you need a real array for whatever reason, you can also spread into one, ie. `[...document.querySelector('.xyz')]`)

src/helpers/image-retry.js
```diff
@@ -0,0 +1,28 @@
...
+    for (let i = 0; i < elements.length; i++) {
+        const img = elements[i];
+        // complete - means it's currently not loading
+        // naturalHeight - will be 0 if loading failed
+        if (!img || !img.complete || img.naturalHeight > 0) continue;
+        // not interested in local URIs
+        if (img.src.startsWith('file://')) continue;
```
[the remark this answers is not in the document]
lostfictions: (Sorry, just saw the icebear PR for unfurling -- didn't get a github notification for some reason. Reading now...)

src/helpers/image-retry.js
```diff
@@ -0,0 +1,28 @@
...
+ *    Note that in a few cases socket reconnect could happen on a live network, but it's not an issue since
+ *    event handler is not heavy.
+ * 2. It finds all image elements, check if they've completed loading and if it was successful
+ * 3. It forces reload of failed images.
+ */
+
+const { socket } = require('~/icebear');
+const { reaction } = require('mobx');
+
+reaction(() => socket.connected, connected => {
```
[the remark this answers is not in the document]
anri-asaturov: Yeah, okay, let's start doing it this way. Even though we have all the other stateful modules.

src/helpers/image-retry.js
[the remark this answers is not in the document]
anri-asaturov: This is just a comment for myself and some others who like to do quick refactor :-D

src/helpers/image-retry.js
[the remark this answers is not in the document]
lostfictions: Small steps!

src/helpers/image-retry.js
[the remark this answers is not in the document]
anri-asaturov: `file://` is what electron does for bundled images, the rest.. like data urls? We have 1 data url currently, the QR code, i'll add `data:`. What else do you think?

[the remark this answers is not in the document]
anri-asaturov: What is 'non-image resource'? You mean 'non image file' and 'non image url'?

src/helpers/network.js
anri-asaturov: Just for clarity. It is safe to call `.open()` and `.close()` on any socket states (although i've noticed a buggy behaviour in case it happens too frequently, that's why `socket.reset()` has a little delay inside).
I just wanted to emphasize when exactly do we want to open and close the socket.

src/helpers/power.js
anri-asaturov: `/auth/push/enable` signals server "don't mind my open connection, i'm not really here"
`/auth/push/enable` signals server "Hey I'm back!"
When socket closes, server know it so we don't need to call '.../enable' on socket close.

anri-asaturov: The API is from mobile that's why naming doesn't make sense

src/ui/chat/components/InlineFiles.jsx
anri-asaturov: Eh, this is a question of 'Do we mind accessibility right now? If we  do - why just alts?'. Linter is set to complain about missing alts and I was lazy to disable it. And well, in this particular case, what is the 'alt' for your avatar that you just chose to upload? 

What do you think we do with accessibility @flohdot @skylarnagao ? Come up with strict rules and validations and task someone with fixing it before enabling those validations?

src/helpers/network.js
lostfictions: Fair enough, but if the condition doesn't matter, I generally prefer removing it for simplicity.

src/helpers/power.js
lostfictions: There's no need to call `enable` on socket close, but is there a need to call `disable` on the socket reopening? That's what I'm asking here. If this `catch` block gets hit and `socket.reset()` is called, `push/disable` will never have actually been sent to the server, right?

src/ui/chat/components/InlineFiles.jsx
lostfictions: Right, I figured it was the linter complaining -- I don't think empty `alt` attributes makes the problem better? :)
We probably just disable the lint globally if we're not using it, or we're going to end up with a satisfied linter and lots of junk attributes that won't actually be helpful if we decide to do an accessibility pass later.

src/helpers/power.js
anri-asaturov: Yes it won't be sent on current connection, because the connection reset makes it irrelevant.
That api changes connection state. But so does disconnection.

lostfictions: Sorry, I don't know if I get it then... push enabling and push disabling implicitly happen on the server on socket disconnect and connect respectively, is that what you're saying?

So is the purpose of the explict call to `push/enable` above to just speed up the enabling of notifications instead of having the server wait for the socket to time out on its end?

In what cases would the `push/disable` socket call above succeed?

src/helpers/power.js
```diff
@@ -4,11 +4,21 @@ const { socket } = require('~/icebear');
...
+    socket.send('/auth/push/enable')
+        .catch(err => {
+            console.error(err);
+            console.log('Failed to enable push notifications on OS sleep.');
+        });
 });
 
 electron.powerMonitor.on('resume', () => {
     console.log('The system is going to resume');
     if (!socket.authenticated) return;
```
lostfictions: I also don't get this line. How is it possible to be authenticated but disconnected?

src/helpers/power.js
lostfictions: I guess what I'm not sure of is why we'd make these API calls instead of just closing the socket on suspend and opening on resume.

flohdot: The original logic is that push notifications will be sent if there are no sockets connected. BUUUT (on a mac) if you put the computer to sleep, the socket will stay connected. So when I close my computer, it could be some arbitrary amount of time until I start getting push notifications. This is similar to when I background the app on Android, and it keeps the socket open. That's what this API was for originally.

That said, yeah, not sure why we'd not just close the connection on sleep.

anri-asaturov: Yes this is way simpler then you are imagining.

1. Normally, server monitors connection state. When disconnected (and some other conditions) - server enables push notifications. When connected - disables push notifications.

2. OS sleep keeps connections kinda alive for some time, but we know that there's no way user receives messages and notification on this device, so we inform server -  `/auth/push/enable`.

3. When OS resumes, if we're still connected - we inform server `/auth/push/enable`, we're back, on same connection.

Now this particular PR takes care of the following case.
- OS resumes
- Connections is alive, so we send `.../disable`
- But it seems that on some low network levels it's not very alive so there's a noticeable lag before ANY data starts flowing in both directions on the same connection, so we wait only 3 seconds for the server to respond and then break connection forcibly.

anri-asaturov: >In what cases would the push/disable socket call above succeed?
when the OS sleep was short it will succeed, will manage to get sent in 3 seconds

src/helpers/image-retry.js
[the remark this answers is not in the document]
lostfictions: I think it makes sense to me, just trying to think through other requests we might make that aren't via sockets. If we have inline media elements at some point I guess we can just reuse most of this logic (except for the `naturalHeight` bit?) And anything that doesn't depend on the DOM will be in icebear instead?

src/helpers/power.js
lostfictions: Okay, I think I get it now, thanks! Wasn't sure under what conditions things would happen.

src/helpers/image-retry.js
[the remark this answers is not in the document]
anri-asaturov: Yes this logic doesn't try to do anything else except reloading failed external images when we already know the url. And it shouldn't really do anything else. And yeah, other media types - can't fix them aforehead :-D

src/helpers/power.js
lostfictions: So it's worth it to keep this logic instead of always closing the socket on sleep because it'll result in better responsiveness if the socket wasn't closed (eg. suspend for a short period on macOS) right?

anri-asaturov: correct

anri-asaturov: Mobile does something similar btw, but subscribes to mobile background events obviously.

src/helpers/image-retry.js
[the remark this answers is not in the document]
lostfictions: :+1: I guess the only other thing I was thinking of was resources via CSS, but Dmitry already mentioned that and you already have it covered :)

src/helpers/power.js
anri-asaturov: @flohdot if we can avoid reconnect - we should, it's an expensive operation

lostfictions: :+1:

src/helpers/image-retry.js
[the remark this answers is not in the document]
anri-asaturov: Actually i wonder if we could extend this to search background-image css props and somehow detect failure. Although that won't be so robust anymore.

src/helpers/power.js
anri-asaturov: It's not possible to be authenticated but disconnected.
This line checks if we can send the '/auth/push/disable' it's in 'auth' namespace so we need an authenticated connection. In other words, if by the time we go the 'resume' event we are not authenticated (and connected or disconnected - we don't care) there's no reason (and posibility) to send '/auth/push/disable'.

anri-asaturov: So there can be 3 cases
- We are authenticated (or at least we think we are, hehe) so we send the message with short timeout
- We are disconnected, so whatever, server doesn't care anymore about our connection state.
- We are connected, but not authenticated yet, so pretty much don't care again, the connection we had when we were sent to sleep is gone.

lostfictions: So this is checking both connection and authentication state? Want to add a comment to that effect?

anri-asaturov: socket.authenticated is used 3 times in desktop and 26 times in sdk, and yes, it's quite logical that you can't be authenticated and disconnected at the same time, not sure why comment it.

lostfictions: Okay. This definitely tripped me up until you explained those three cases and I realized that this statement covers cases 2 and 3, but probably I'm just slow and everyone else gets it

anri-asaturov: I'll add my cases explanation to the comment then, at least it makes sense compared to going to all 29 uses of socket.authenticated and writing `// if you're authenticated you're also connected ` :-D

src/helpers/image-retry.js
```diff
@@ -11,18 +11,30 @@
...
-        // complete - means it's currently not loading
-        // naturalHeight - will be 0 if loading failed
-        if (!img || !img.complete || img.naturalHeight > 0) continue;
-        // not interested in local URIs
-        if (img.src.startsWith('file://')) continue;
-        // magically making image reload
-        img.src = img.src;
-    }
-});
+let disposer;
```
lostfictions: Sorry, this isn't what I meant :(  This is still a singleton and is even more stateful now, and I don't think it improves clarity or maintainability compared to your previous version.

The reason I was suggesting wrapping it in a function was to pass in its dependencies instead of relying on those other modules also being singletons. So it could look something like this:
```js
function retryImagesOnReconnect(socket) {
    return reaction(() => socket.connected, connected => {
        if (!connected) return;
        // caution: this is not an array, no forEach is available
        const elements = document.getElementsByTagName('img');
        for (let i = 0; i < elements.length; i++) {
            const img = elements[i];
            // complete - means it's currently not loading
            // naturalHeight - will be 0 if loading failed
            if (!img || !img.complete || img.naturalHeight > 0) continue;
            // not interested in local URIs
            if (img.src.startsWith('file://')) continue;
            // magically making image reload
            img.src = img.src;
        }
    });
}
```
and then there's no need to depend on there being a single socket that exists globally in the module. Does that make sense? Do you disagree that this is worthwhile?

lostfictions: Ditto below -- you'd pass both the `appState` and the `socket` instead of importing them. IMO making these dependencies explicit is a good general pattern and improves testability and maintainability quite a lot, even if in these cases these methods are likely only going to be called once in one place. I pretty strongly feel like it's a good idea to start moving towards this kind of structure whenever possible even though I know a lot of our codebase isn't built like this.

src/helpers/image-retry.js
```diff
@@ -11,18 +11,30 @@
...
+
+function start() {
+    if (disposer) return;
+    disposer = reaction(() => socket.connected, connected => {
+        if (!connected) return;
+        // caution: this is not an array, no forEach is available
+        const elements = document.getElementsByTagName('img');
+        for (let i = 0; i < elements.length; i++) {
```
lostfictions: Is there a reason to keep this like this instead of using `for (const element of ...` like I mentioned?

src/helpers/image-retry.js
lostfictions: One immediate benefit this would provide is not having to create more "DI modules" (a pattern I've never actually seen before) like
https://github.com/PeerioTechnologies/peerio-icebear/blob/dev/src/helpers/di-chat-store.js
https://github.com/PeerioTechnologies/peerio-icebear/blob/dev/src/helpers/di-contact-store.js
https://github.com/PeerioTechnologies/peerio-icebear/blob/dev/src/helpers/di-current-user.js
https://github.com/PeerioTechnologies/peerio-icebear/blob/dev/src/helpers/di-file-store.js
etc.

src/helpers/image-retry.js
anri-asaturov: Other then it's very slightly slower? Probably no, not in this case for sure. But it's also not such a big of an improvement.

anri-asaturov: I'll try to use it more, that said.

src/helpers/image-retry.js
anri-asaturov: - I do agree that eventually we need to refactor repos to achieve some sensible level of decoupling 
- I don't think that time is now and especially in this PR

We also kinda need to agree on how are we gonna do it first, cuz passing dependency as an argument might not be the DI of our choice. And back to typescript and tests - we probably need those first, decouple right after?

anri-asaturov: Can you tell me of your preferred ways to organize decoupled modules and DI? We can document the decision and start writing new modules properly and maybe refactoring some small parts even before switching to typescript and having more tests?

lostfictions: TBH I don't think of wrapping a side-effectful thing in a function and passing its operand like this as "DI"? To me it's just a prerequisite and a building block to doing anything more complex, but in itself it's still very useful -- and a simple, low-hanging fruit to not increase entropy in the codebase.

It's not a disruptive, all-encompassing change that needs types and tests; to me it's just good best practice and generally a no-brainer, which is why I suggested it in this PR. I think it's a helpful approach moving forward for new modules with no complex dependencies like this one, regardless of any planned further refactors to existing code. But okay, i guess this is already merged so whatever.

anri-asaturov: Why not? Module has a dependency and we're passing it to the module, looks like DI to me :-D

So what if later we decide we pass a special object to every module and dependencies are in there. Or maybe we'll make angular-like DI (i hope not). Or something else. Let's decide now(in next few days) on the pattern and start implementing it?

PR #153, opened 2017-11-27
Enable Windows updater

dchest:
* Electron-updater removed 🎉 

* This PR also bumps peerio-updater version, which contains an unrelated fix for Linux updater (https://peerio.tpondemand.com/entity/7621). I though it would be okay to include it here, since it would be annoying to make separate PRs for that, and since both PRs would have to be tested via two staging releases.

Not sure we have a task for it.

For testing on staging we have these:
https://peerio.tpondemand.com/entity/6163

It was already enabled on staging for some time, so we tested it. However, we should do the usual two-staging releases test:

1. Release staging.
2. Check that it launches.
3. Release another staging.
4. Check that the previous one updates to it.

Let's test all platforms.

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


anri-asaturov:
@ebilg please take testing instructions into consideration, we need to test this week's RC as I will merge this now.

dchest:
Indeed, since we also have this merged https://github.com/PeerioTechnologies/peerio-desktop/pull/142 which requires Mac updater testing, we need to test updates on all platforms.

PR #160, opened 2017-11-29
feat: UI feedback for poor connection when downloading URL image

suprko:
[PR guideline](https://github.com/PeerioTechnologies/peerio-icebear/blob/dev/docs/CONTRIBUTING.md)

#### Relevant issue tracker links  
https://peerio.tpondemand.com/entity/7654
#### Linked PRs in other repos
n/a
#### Manual testing instructions  
"Poor connection..." message should appear when client app is attempting to download URL image preview, but the app is offline. It should not appear if: image is not supposed to download (e.g. oversize or user has disabled image previews), or if user is online.
#### Additional notes
Not related to this particular task, but I'm finding that the live retrying of images downloads isn't working. I need to click to a different chat and then come back.

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Big feature tech design documentation.
* Readability.
* DRY.
* Can we extract something to SDK?
* Are there any bicycles invented that should be replaced with existing SDK
* Are there unit and e2e tests added?

----
### Repository owner

- [ ] At least one approved code review exists.
- [ ] Approved tester review exists and was done after last code changes.
- [ ] CI build and tests passed.
- [ ] This feature/bugfix is considered fully done and can get merged.
- [ ] PR name follows conventional changelog format.
- [ ] I'll remember to squash commits.

PR #161, opened 2017-11-29
feat: UI feedback for poor connection when downloading inline image preview

suprko:

https://peerio.tpondemand.com/entity/7652
PR for similar functionality for URL previews:
https://github.com/PeerioTechnologies/peerio-desktop/pull/160
"Poor connection..." message should appear when client app is trying to download an inline image preview, but app is offline. It should not appear if: image is not supposed to download (e.g. oversize, or user has disabled previews), or if user is online.
#### Additional notes
The mockups show the progress bar tracking with how much of the file has been downloaded. My experience (on pretty fast internet) is that the download is so quick that you almost never see the download bar progress, just a very quick flash right before the download completes and the image shows up. So maybe it would be better to switch this to the "indefinite" style progress bar, as is used for URL previews.

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


src/ui/chat/components/InlineFiles.jsx
```diff
@@ -297,13 +298,16 @@ class InlineFile extends React.Component {
...
                                                 onError={this.onErrorLoadingImage}
-                                                onClick={this.imageClick} />)}
+                                                onClick={this.imageClick} />) }
                                     {!file.tmpCached && !file.downloading && file.isOverInlineSizeLimit
                                         && !file.isOversizeCutoff
                                         && this.renderOversizeWarning()}
                                     {file.isOversizeCutoff
                                         && this.renderOversizeCutoffWarning()}
                                     {file.cachingFailed ? <span>{t('error_downloadFailed')}</span> : null}
+                                    { file.downloading && !appState.isOnline &&
```
anri-asaturov: rely on `socket.authenticated` instead of `appState.isOnline` please here. The reason is that you might be online or even connected, but if you're not authenticated you're for sure not downloading anything.

anri-asaturov: There's another problem though is that this warning will be shown just for a few seconds until it falls into `cachingFailed`

suprko: OK, switched to `socket.authenticated`.

Re: `cachingFailed`, I agree, in general I need more detail from design on this... Do we still want the download attempt to timeout and show "Download failed", or does the new message replace "Download failed"? Do we want "poor connection" to show up immediately when user app is offline or is there some timeout? I'll keep an eye on your TP comment.

anri-asaturov: Yeah i can merge this and it technically works, but it's weird. Download failed, but will retry on reconnect..

anri-asaturov:
reopen this when/if needed

PR #165, opened 2017-12-01
fix: small bugs

gembarrett:

### PR author
Gem

https://peerio.tpondemand.com/entity/6297
https://peerio.tpondemand.com/entity/6783

... 
...
#### Additional notes
... 

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


src/ui/chat/components/ChatList.jsx
```diff
@@ -158,7 +158,7 @@ class ChatList extends React.Component {
                                             }
                                             itemContent={
                                                 <TooltipDiv className="item-content"
-                                                    tooltip={c.name}
+                                                    tooltip={c.participantUsernames[0]}
```
anri-asaturov: Chat with self has 0 elements in this array. Check what it renders, if empty string - fine i guess, but it could render something like 'undefined'

anri-asaturov:
@gembarrett ci fails, fix plz. btw you can create `.opt-in` file  with `lint` in it and it will run eslint on commit.

Waiting for @ebilg .

flohdot:
lgtm

src/ui/signup/AvatarControl.jsx
```diff
@@ -6,16 +6,21 @@ const AvatarDialog = require('./AvatarDialog');
...
     render() {
         const { url } = this.props;
         return (
             <div className="avatar-input" onClick={AvatarDialog.showDialog}>
-                <FontIcon value="add" />
-                <div className="avatar-instructions">
-                    {t('title_avatarInstructions')}
-                </div>
+                {!url && this.getInstructions()}
                 {url && <img src={url} />}
```
suprko: If these two are dependent on the same variable `url` then it's cleaner to do this as `{url ?  : this.getInstructions() }`

src/ui/signup/AvatarControl.jsx
```diff
@@ -6,16 +6,21 @@ const AvatarDialog = require('./AvatarDialog');
...
+    }
     render() {
         const { url } = this.props;
         return (
             <div className="avatar-input" onClick={AvatarDialog.showDialog}>
-                <FontIcon value="add" />
-                <div className="avatar-instructions">
-                    {t('title_avatarInstructions')}
-                </div>
+                {!url && this.getInstructions()}
```
suprko: (sorry for the double comment. this one is inline with the code now)

If these two are dependent on the same variable `url` then it's cleaner to do this as `{url ?  : this.getInstructions() }`

anri-asaturov:
Still waiting on @gembarrett to answer

src/ui/chat/components/ChatList.jsx
gembarrett: Good catch! As a fallback should c.name be used in the case of an empty/undefined string @flohbot ?

anri-asaturov:
```    @computed get name() {
        if (this.isChannel && this.chatHead && this.chatHead.chatName) return this.chatHead.chatName;
        return this.participants.length === 0
            ? (User.current.fullName || User.current.username)
            : this.participants.map(p => p.fullName || p.username).join(', ');
    }```
This is how chat.name is getting generated, so if you use it, you'll get the full name of current user.

Comment (flohdot) [reply src/ui/chat/components/ChatList.jsx]: I think an empty string is fine actually. Or maybe `t('title_you')` so that it just says "you"

Comment (flohdot) [reply src/ui/chat/components/ChatList.jsx]: as fallback, that is to say

File: src/ui/signup/AvatarControl.jsx (hunk 3 above)
Comment (gembarrett) [reply src/ui/signup/AvatarControl.jsx]: Great, thanks

File: src/ui/signup/AvatarControl.jsx [hunk 4]
```diff
@@ -6,16 +6,20 @@ const AvatarDialog = require('./AvatarDialog');
 const { t } = require('peerio-translator');
 
 @observer class AvatarControl extends Component {
+    getInstructions() {
+        return (
+            <div className="avatar-instructions">
+                <FontIcon value="add" />
+                <p className="avatar-instructions">{t('title_avatarInstructions')}</p>
```
suprko: We've got a p class `avatar-instructions` nested inside a div class `avatar-instructions`! Is this deliberate?

gembarrett: Oh oops! Must've forgotten to delete one, will sort now

PR #166, opened 2017-12-01
chore: package.json improvements

lostfictions:
tried to run the build on a machine where it was not ideal to set the socket server in a shell rc script. I made a script to set it before running npm start, but it's going to get erased if i accidentally checkout or reset -- i thought as a stopgap to migrating to dot-env/envalid we could just make staging the default for the default run command.

this pr also reorganizes the scripts to use the "task:target" format instead of "target:task" -- this lets us simplify npm-run-all commands to their use the less verbose and more maintainable `compile:*` shorthand (see [here](https://github.com/mysticatea/npm-run-all/blob/master/docs/run-p.md#glob-like-pattern-matching-for-script-names)).


anri-asaturov:
👍

PR #171, opened 2017-12-04
feat: after downloading file, show it in folder instead of previewing it

dchest:
Files are no longer previewed after downloading, instead the destination directory is opened with system file manager with the downloaded file selected.

...

https://github.com/PeerioTechnologies/peerio-icebear/pull/88

1. Download some file.
2. Make sure it opens in system file manager after downloading with the file selected instead of showing its preview.

#### Additional notes
... 

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


src/helpers/file.js
```diff
@@ -3,6 +3,16 @@ const fs = require('fs');
 const _ = require('lodash');
 const path = require('path');
 const sanitize = require('sanitize-filename');
+const util = require('util');
+
+function renameFile(oldPath, newPath) {
+    return new Promise((fulfill, reject) => {
+        fs.rename(oldPath, newPath, err => {
```
anri-asaturov: why not renameSync?

dchest: Don't wanna block, e.g. if spinning HDD has bad sectors and is trying to find a good one :)

dchest: Once `util.promisify` is in Electron, we can replace this with `renameFile = util.promisify(fs.rename)`.

anri-asaturov: How to detect an old man - he knows what 'bad sectors' are :-D
fs promisify is better to be done on a global level, so don't bother for now.

dchest: Haha. Well, I believe filled SSDs/flash drives still have delays when their controller is trying to find a place to write.

dchest:
BTW, we should delete the .peeriodownload file if downloading files, right? Not sure how resume works currently or if it works at all.

anri-asaturov:
Oh, crap, this will break resume. You have to put it in sdk.

anri-asaturov:
Or maybe it won't break, let Eren test it first.

dchest:
Yup, it will break. So no tests for now :)

dchest:
I moved test steps to https://github.com/PeerioTechnologies/peerio-icebear/pull/89 and this PR is now only about opening folder for downloaded file instead of previewing it.

PR #174, opened 2017-12-05
feat(notifications): focus on input when clicking notification

dchest:

https://peerio.tpondemand.com/entity/7720

... 

0. Put focus outside of message input (e.g. click on messages)
1. Click on notification.
2. Check that message input is focused.

#### Additional notes
... 

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


lostfictions:
The `onBlur` is always called when unmounting? Should `uiStore.messageInputFocused` be set to `false` when the message input component is unmounted regardless, just to make sure it's in a consistent state?

Setting `messageInputFocused` to `false` when the component _is_ mounted won't blur it, also -- probably not a big deal, but might surprise someone later.

anri-asaturov:
A bit too complex to my taste. I would make 
``` uiStore.focusMessageInput(){ if (this.inputRef) inputRef.focus()}```
And then when message input mounts/unmounts i would update inputRef in the uiStore

**Comment by anri-asaturov** [comment]
My motivation is that messageInputFocused is not really a state, just a way to pass an event.

**Comment by dchest** [comment]
Yeah, I like your suggestion. Input focus isn't really good with declarative-style state management.

**Comment by dchest** [comment]
Please re-review.

**Comment by lostfictions** [comment]
i do sort of wish we had some better mechanism for having actions flow down that didn't require us to keep extra state like this, but that's not really a comment on this PR :)

**Comment by dchest** [comment]
@lostfictions true

File: src/ui/chat/components/MessageInputProseMirror.jsx [hunk 1]
```diff
@@ -305,6 +306,7 @@ class MessageInputProseMirror extends React.Component {
     mountProseMirror(el) {
         if (!el) return;
```
anri-asaturov: we should do `uiStore.messageInputEditorView = null` in here

src/stores/ui-store.js
```diff
@@ -28,6 +28,15 @@ class UIStore {
     // message drafts, not persisted. key: chat id, value: text
     unsentMessages = {};
 
+    // Reference to ProseMirror message input editor.
+    messageInputEditorView = null;
```
anri-asaturov: Just so we don't get completely unholy, let's reference our react component in here, and not the ref to actual prosemirror element. Also the `focus()` in our component should be protected with try catch just in case.

dchest: ok

lostfictions:
@anri-asaturov if you're going to ask @dchest to build extra wrapping abstraction around this, why don't we just put an [Atom](https://mobx.js.org/refguide/extending.html) in the UIStore and have the input component set up a `reaction`?

anri-asaturov:
Why not, I'm always pro-torturing Dmitry.

lostfictions:
I really didn't want to bikeshed for such a small change, but it might be useful to set up a small observable abstraction for these kinds of cases -- something more like a traditional/streams/Rx-style observable, that can be subscribed to by components and invoked by stuff like the notification click.

anri-asaturov:
EventEmmiter?

lostfictions:
Do we have an abstraction for those that's observable?

anri-asaturov:
Nope, I'd love to have something.

anri-asaturov:
@dchest sorry we confused you with so many options, just wrap the .focus() with try{}catch just in case and not that if 'ref' function gets undefined argument - the ref doesn't exist anymore.

anri-asaturov:
@ebilg this can be tested

ebilg:
focus working correctly

dchest:
@anri-asaturov 

> Just so we don't get completely unholy, let's reference our react component in here, and not the ref to actual prosemirror element.

I wanted to do it initially, but figured that the code in notifications.js already doesn't care what kind of thing it calls .focus() on as log as the method exists, and if we gonna change something in message input component, we can as well change the reference to be the React component and add .focus() method in it.

@lostfictions 

> I really didn't want to bikeshed for such a small change, but it might be useful to set up a small observable abstraction for these kinds of cases -- something more like a traditional/streams/Rx-style observable, that can be subscribed to by components and invoked by stuff like the notification click.

It seems to me we should stick to keeping state/stores as a single source of the truth for UI as much as possible, which is why my first attempt was using it, but it wasn't really working for focus. In theory, such things like focus, which are hard to manage as external state (from DOM), are pretty uncommon. I agree though that even these uncommon situations need some architectural design, so maybe it's worth trying what you proposed. If we have (or *when* we have) something else that needs this, we can refactor both. My only fear is that if we add events/streams, people will start using them for things that should be in the state, and then we'll have a mess of the state. But we can catch this during review.

flohdot:
Mobile is gonna have to implement something similar (but... worse) with push notifications soon, so maybe let's keep thinking about what the correct pattern is so that we can make recommendations to @seavan or whoever ends up working on this regarding the architecture. I agree that this has some potential for chaos.

PR #176, opened 2017-12-11
fix: on Windows use %LOCALAPPDATA% instead of default %APPDATA% for storing local files

dchest:
The default path was storing files in the Roaming directory. We changed
it to Local. We also migrate from the old place to the new one by moving
the whole directory if the new directory doesn't exist.

https://peerio.tpondemand.com/entity/7785 
https://app.clubhouse.io/peerio/story/712/desktop-windows-user-data-folder-is-in-roaming

#### Testing instructions  

Testing is required on Windows only (Win7 too please).

A.

1. Before starting the app, in Windows Explorer open `AppData\Roaming` for the user and check that "Peerio 2" folder is there, and in another window `AppData\Local` and notice that "Peerio 2" folder is *not* there (this things can be typed right into the path bar).

2. Start the app.

3. Notice that "Peerio 2" folder is moved from `AppData\Roaming` to `AppData\Local` (if not, it's a bug).

4. Quit the app and start it again.

5. Check that "Peerio 2" folder is still in `AppData\Local` and there's no such folder in the `AppData\Roaming`.

Also, we should thing about failures that can happen. (If the folder fails to move for some reason, Peerio will continue using the old `Roaming` folder.)

B.

On a "freshly installed app", which doesn't have the corresponding data folder in `AppData\Roaming` or `AppData\Local`, make sure when the app starts, the data is saved to `AppData\Local` folder, not the roaming one.

----
### Repository owner


src/config.js
```diff
@@ -8,6 +9,44 @@ const StorageEngine = require('peerio-icebear/dist/models/storage/node-json-stor
...
+            // XXX: moving can be removed in the future when older versions
+            // of the client are deprecated and everyone's updated.
+            if (!fs.existsSync(newUserData)) {
+                const fse = require('fs-extra');
+                fse.moveSync(oldUserData, newUserData);
+                console.log(`Moved ${oldUserData} to ${newUserData}`);
+            }
+
+            // Set new paths.
+            app.setPath('appData', process.env.LOCALAPPDATA);
```
anri-asaturov: https://github.com/PeerioTechnologies/peerio-desktop/blob/85bdf0e775ded4ed74bc5a3aa3cd944b39ac1c7a/src/main.js#L18

eh?

dchest: right, I'll move this to config.js

dchest: done

src/config.js
```diff
@@ -8,6 +9,49 @@ const StorageEngine = require('peerio-icebear/dist/models/storage/node-json-stor
...
+        } catch (e) {
+            // Failed to move old folder, so continue using it.
+            console.error(e);
+        }
+    }
+}
+
+// For dev builds we want to use separate user data directory
+if (isDevEnv) {
+    app.setPath('userData', path.resolve(app.getPath('appData'), `${app.getName().toLowerCase()}_dev`));
```
anri-asaturov: config.js executes once in main process and once in renderer, this code won't work as expected.

dchest: Ah, true. I'll move all of this into the main.js, right?

anri-asaturov: That should work i think

dchest: done

src/main.js
```diff
@@ -12,11 +12,52 @@ process.on('uncaughtException', (error) => {
...
+            // all our old data to into a new directory if the new userData
+            // directory doesn't exist.
+            //
+            // This needs to happen before Electron's app 'ready' event.
+            // Also, everything needs to be synchronous.
+            //
+            // XXX: moving can be removed in the future when older versions
+            // of the client are deprecated and everyone's updated.
+            const fse = require('fs-extra');
+            if (!fse.existsSync(newUserData)) {
```
anri-asaturov: 1. What if it's a fresh install?
2. Are you sure that under no circumstances oldUserData doesn't  equal to something global like `C:\Users\anri\AppData\Roaming` and we won't  be moving that?

dchest: 1. Fixed.
2. Yes, see https://github.com/electron/electron/blob/master/lib/browser/init.js#L149

Now thinking if it's possible to get the folder name without environment variables (and binary modules)...

anri-asaturov:
ready for testing

ebilg:
working on Windows 7 and 10

PR #188, opened 2017-12-19
fix: window sizing saving and cache deletion on quit

dchest:
We were lucky that window sizing saving code managed to save before the app quit (we didn't wait for the promise). I fixed that, and also moved cache deletion into `will-quit` hook.

https://app.clubhouse.io/peerio/story/1212/consider-moving-cache-deletion-to-app-quit-hook
 
#### Testing instructions  

Since this code listens to `will-quit` hook, and Peerio Updater also listens to it, I want to make sure that it won't interfere with updater (which is unlikely). So, @anri please release two staging RC 
builds for the release instead of the usual one, after this is merged so that we could test that updates work.

These can be tested without staging releases:

To test cache deletion: after quitting the app, check that the cache folder is gone. Quitting the app can be done in two ways: Cmd+Q and closing the main window (both paths should be tested).

To test window restore: resize the window, quit, launch again and see if the size is preserved.

----
### Repository owner


anri-asaturov:
@ebilg ready to test

dchest:
Sorry, fixed one tiny thing :)

ebilg:
looks good

PR #190, opened 2017-12-20
fix: file picker scroller querySelector

seavan:
https://app.clubhouse.io/peerio/story/1351/desktop-share-from-files-to-chat-dialog-only-shows-first-page-of-files


seavan:
@ebilg tested, says looks good

anri-asaturov:
It's one landmine replaced with another landmine, i'll fix it myself.

PR #193, opened 2017-12-20
fix: unbreak KegEditor's SELF display

dchest:
#### Testing instructions  

----
### Repository owner


src/ui/dev-tools/KegEditor.jsx
```diff
@@ -147,7 +147,9 @@ function ChatInfo(props) {
         <div className="selectable">
             <Chip className="selectable">{props.c.id}</Chip><br />
             {
-                props.c.allParticipants.map(p => <Chip key={p.username}>{p.username}</Chip>)
+                props.c.allParticipants
```
anri-asaturov: Chat#allParticipants is never null, are u sure this fixes something?

dchest: Yes, it's not chat when it's self.

dchest: Broke here https://github.com/PeerioTechnologies/peerio-desktop/commit/f67bd255402382fb57901fa61b7e94f5a2d4e6e9#diff-cf0c8119d86e68d523c28eee8a68a0b7

dchest: I guess the component name is misleading.

PR #194, opened 2017-12-20
chore: replace RTB components Link, Card, Checkbox, FontIcon

suprko:
https://app.clubhouse.io/peerio/story/1272/desktop-replace-card-and-link-components
https://app.clubhouse.io/peerio/story/1273/desktop-replace-checkbox-components
https://app.clubhouse.io/peerio/story/1285/desktop-replace-fonticon-component

#### Testing instructions  
This is one of those annoying `chore`s where a lot of things are affected but in a really minor way.

1. All `Card` and `Link` RTB components have been replaced with plain markup. There were only a few no need to test as they're extremely simple and I've verified that they look/work as they should.
2. All `Checkbox` RTB components have been replaced with our own `Checkbox` component. There was one `Button` that was pretending to be a checkbox that has also been converted. Here are their locations:
  * 2FA login dialog
  * logout "untrust device" dialog
  * URL preview "only from favourites"
  * "Share from my files" dialog (multi-select files)
3. All `FontIcon` RTB components have been replaced with our own `MaterialIcon`. This is every instance of an icon that is *not* in a `Button`, `ListItem`, `MenuItem`... Basically this affects any instance of a Material icon that is used purely decoratively and is not clickable/interactable. While these are kind of everywhere through the app, I don't think it needs to be tested intensively, due to the lack of interactivity. I gave each page a once-over when I updated it so IMO it's fine.

----
### Repository owner


suprko:
Linking to Clubhouse: [ch1272] [ch1273] [ch1285]

anri-asaturov:
wow wow, easy :-D

suprko:
[ch1272]
[ch1273]
[ch1285]

src/ui/files/components/FileLine.jsx
```diff
@@ -107,12 +111,10 @@ class FileLine extends React.Component {
...
 
                 {this.props.checkbox &&
-                    <Button
-                        className="checkbox no-ripple"
-                        icon={this.props.selected
-                            ? 'check_box'
-                            : 'check_box_outline_blank'}
-                        onClick={this.props.onToggleSelect}
+                    <Checkbox
+                        className="checkbox"
```
anri-asaturov: can this go to the component?

src/ui/dev-tools/Dashboard.jsx
```diff
@@ -9,24 +9,18 @@ class DevToolsDashboard extends React.Component {
...
-                        <UserDetails />
-                    </CardText>
-                </Card>
-                <Card style={{ width: '57%' }}>
-                    <CardTitle title="Config" />
-                    <CardText>
-                        <pre className="selectable">{JSON.stringify(config, null, 2)}</pre>
-                    </CardText>
-                </Card>
+                <div className="card">
```
anri-asaturov: We don't want a Card component?

anri-asaturov: The good thing about having a component is that you don't need to remember class names.

suprko: No problem making a Card component if we want! I didn't make one because it's only used in 1 place (dev tools)

anri-asaturov: right, nvm

src/ui/files/components/FileLine.jsx
suprko: So this part is confusing... We refactored Files view to no longer use `<table>`, but as a result we had to do some hacky things to make sure it's still in a "grid".

One of the things we had to do is put a checkbox placeholder, an empty div that renders when a `FileLine` or `FolderLine` does *not* have a checkbox, which is the same width and has the same margins as the real checkbox. These are all marked with a `.checkbox` class, and since we want the placeholder and real checkbox to have the same sizing and spacing, we give them both the same class.

Maybe I can rename it `files-checkbox` so it's more obvious that it's specific to this view and not something needed for all `<Checkbox>`s?

anri-asaturov: yes, please

suprko:
@anri-asaturov forgot to push yesterday :man_facepalming: updated now

PR #196, opened 2017-12-22
chore: replace RTB <Dropdown> and <RadioGroup> components

suprko:
https://app.clubhouse.io/peerio/story/1332/desktop-replace-dropdown-component
https://app.clubhouse.io/peerio/story/1336/desktop-replace-radio-components

#### Testing instructions  
These components don't appear much in app.
* Dropdown is used in the ChatLlist "sort by" thing
* RadioGroup used in the URL image consent and Peerio image consent prompts

In my quick tests they were all working, may or may not be worth passing to Eren as well.

----
### Repository owner


src/ui-library/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+            this.isActive = false;
+        }
+    }
+
+    render() {
+        const { value, options } = this.props;
+        const dropdownOptions = [];
+        const labels = {};
+
+        for (let i = 0; i < options.length; i++) {
```
anri-asaturov: Don't run this cycle when dropdown is not active

src/ui-library/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+        this.props.onChange(ev.target.getAttribute('data-value'));
+        this.isActive = false;
+    }
+
+    // Make the dropdown close when you click away
+    @observable isHover = false;
+    @action.bound handleMouseEnter() { this.isHover = true; }
+    @action.bound handleMouseLeave() { this.isHover = false; }
+
+    componentDidMount() { window.addEventListener('click', this.handleClick, false); }
```
anri-asaturov: Ok lines 32-33 i can understand, although also not preferred, but this shouldn't be a one-liner

src/ui-library/Dropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+                >
+                    <div className="selector clickable" onClick={this.activate}>
+                        <span className="input">
+                            {labels[value]}
+                        </span>
+                        <span className="material-icons">
+                            arrow_drop_down
+                        </span>
+                    </div>
+                    <ul className={this.isActive ? 'active' : null}>
```
anri-asaturov: u can just not render the ul instead of setting class, unless you want to animate it

src/ui-library/Dropdown.jsx
suprko: Hmmm not sure what to do here because the `labels` object is needed to show the currently selected value. We could have something like a `valueLabel` prop, but then a loop like this would need to be done in the parent component, so in the end that wouldn't save us anything...

anri-asaturov: Add only selected item to the arrays

suprko: If I understand this right: when the component loads, the currently selected item is in the array, and then when you open the dropdown the rest of the items get added to the array, which means that the items in the dropdown menu will be in a different order depending on which item was initially selected...

PR #201, opened 2018-01-04
WIP: replace RTB tooltip, avatar components

suprko:
https://app.clubhouse.io/peerio/story/1473/desktop-create-alternative-for-tooltip-components
https://app.clubhouse.io/peerio/story/1330/desktop-replace-avatar-component

#### Testing instructions
All avatars and tooltips should work the same as they have before.

Currently WIP, I'll keep track of where everything has been replaced in these lists:

**Avatars** (with and without tooltips)
* AppNav (self avatar, far left sidebar)
* ChatList (left sidebar)
* Chat
  * top of Room or DM (list of users)
  * read receipts
  * left of each message
* MemberSection (right sidebar of a Room)
* MessageSideBar (right sidebar after clicking read receipts)
* message input - user suggestions when you start typing "@"
* Contacts > Favourites
* Contacts > All Contacts

**Tooltips**
* (none so far)

----
### Repository owner


suprko:
@anri-asaturov @seavan looking to get some comments on the components themselves before I go through the entire app and replace everything

src/ui-library/Avatar.jsx
```diff
@@ -0,0 +1,96 @@
...
+    ----------------------------------------
+    className       string
+    size            string      small, medium (default), large
+    clickable       bool        default false
+    tooltip         bool        default false
+
+    contact         string      *
+    username        string      * need one of contact or username
+    ----------------------------------------
+*/
```
seavan: we should switch to PropTypes for this kind of thing, it also works well with VSCode code completion.

you could add comments to proptypes lines:

https://reactjs.org/docs/typechecking-with-proptypes.html

src/ui-library/Avatar.jsx
```diff
@@ -0,0 +1,96 @@
...
+        this.coordinates = getPositionInWindow(ev.target);
+        this.hovered = true;
+    }
+
+    @action.bound handleMouseLeave() {
+        this.hovered = false;
+    }
+
+    // When avatar is clickable, click opens ContactProfile dialog
+    openContactDialog = (ev) => {
```
seavan: now we use @action.bound everywhere, for uniformity

src/ui-library/Tooltip.jsx
```diff
@@ -0,0 +1,91 @@
...
+        Then opacity:0 removed via isVisible=true.
+    */
+    @observable isVisible = false;
+    style = {};
+
+    componentDidMount() {
+        this.calculateDimensions();
+    }
+
+    calculateDimensions = () => {
```
seavan: no reason for arrow, this function is only called from componentDidMount, meaning "this" is always correct.

src/ui-library/Avatar.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+    contact         string      *
+    username        string      * need one of contact or username
+    ----------------------------------------
+*/
+
+@observer
+class Avatar extends React.Component {
+    @observable hovered = false;
+    @observable coordinates = {};
```
anri-asaturov: It doesn't have to be changed, but just so you understand what's happening in here:
You create an observable property and assign a new empty object to it, this empty object's properties will never be observable, but since you replace default object - it works, this assignment is just a no-op.

src/ui-library/Avatar.jsx
```diff
@@ -0,0 +1,96 @@
...
+                            ? <img src={c.mediumAvatarUrl} alt={c.username} />
+                            : c.letter
+                        }
+                    </div>
+                </div>
+                {this.props.tooltip && this.hovered
+                    ? <Tooltip
+                        text={c.fullNameAndUsername}
+                        position="top"
+                        coordinates={this.coordinates}
```
anri-asaturov: Tooltip receives coordinates? Noooo, this will make it very hard to use, since every consumer will have to deal with coordinate detection, this is 100% Tooltip's job

anri-asaturov: You can pass container `ref` or `id` or something else that will help you figure out the container. Maybe even navigate parent chain to find an element with 'tooltip-container' attribute or something

src/ui-library/helpers.js
```diff
@@ -0,0 +1,21 @@
...
+ * @param {Object} element - HTML element
+ * @returns {Object}
+ */
+function getPositionInWindow(element) {
+    const { width, height } = element.getBoundingClientRect();
+
+    return {
+        offsetX: width / 2,
+        offsetY: height / 2,
+        posX: element.getBoundingClientRect().left + (width / 2),
```
anri-asaturov: you just called this function at line 9 and extracted width and height, why not do the same for left and top

src/ui-library/Tooltip.jsx
```diff
@@ -0,0 +1,93 @@
...
+            (It ~should~ only be possible to have one tooltip onscreen at a time.)
+        */
+        let tooltip;
+        if (this.props.id) {
+            tooltip = document.getElementById(this.props.id);
+        } else {
+            tooltip = document.getElementsByClassName('p-tooltip')[0];
+        }
+
+        const tooltipContainer = tooltip.parentElement;
```
suprko: @anri-asaturov do you think this is alright -- just go to the `parentElement`? Or should it be more specific, check for `data-tooltip-container` or something?

src/ui-library/Avatar.jsx
suprko: Do you think it's worth implementing this if we're planning to move to TypeScript anyway?

src/ui-library/Tooltip.jsx
anri-asaturov: It's good

suprko:
@anri-asaturov sorry for all the title changes, this is in fact still WIP because I haven't replaced all of the tooltip components thru the app

suprko:
@anri-asaturov arg, I didn't think this through... I won't actually be able to replace the rest of the tooltips without first replacing those components themselves (Button, Menu, etc.). Did this backwards.

suprko:
https://github.com/PeerioTechnologies/peerio-desktop/pull/205

PR #205, opened 2018-01-11
chore: replace RTB <Button>, <Avatar>, <Tooltip...>

suprko:
#### Relevant info and issue/PR links 
 https://app.clubhouse.io/peerio/story/1331/desktop-replace-button-component
https://app.clubhouse.io/peerio/story/1330/desktop-replace-avatar-component
https://app.clubhouse.io/peerio/story/1473/desktop-create-alternative-for-tooltip-components

#### Testing instructions  
OK. Massive PR that affects almost every view in the entire app. I promise after this they get much easier. This PR replaces the RTB versions of:

* every Avatar
* every Button (not including those in Dialogs) and TooltipIconButton
* every TooltipDiv

I tested each view one-by-one as I replaced the components, but it would still be good to have eyes on them.

----
### Repository owner

- [ ] Was tested and can be merged.
- [ ] PR name follows conventional changelog format.

[PR guideline](https://github.com/PeerioTechnologies/peerio-icebear/blob/dev/docs/CONTRIBUTING.md)

PR #207, opened 2018-01-12
chore: replace RTB <Dialog>, <IconButton>

suprko:
https://app.clubhouse.io/peerio/story/1333/desktop-replace-dialog-component
(also some leftover `Button` replacements; I hadn't replaced the `IconButtons`!)

#### Testing instructions  
Similar to https://github.com/PeerioTechnologies/peerio-desktop/pull/205 (but not nearly as big), this is a PR that affects a lot of little things in the app. All `IconButton`s have been replaced (they were missed on the previous PR), and all `Dialog`s as well.

----
### Repository owner


suprko:
@anri-asaturov looking for feedback on how I've set up `Dialog.jsx`. Specifically using `reaction` and `setTimeout` to make sure that the fade in/out transition plays. I tried to use `ref` for this, but I could only get that to work on component mount, not unmount.

suprko:
^ and @seavan !

src/ui-library/Dialog.jsx
```diff
@@ -0,0 +1,118 @@
...
+        */
+        this.activeReaction = reaction(() => this.props.active, active => {
+            if (active) {
+                this.dialogRendered = true;
+                setTimeout(() => {
+                    this.dialogVisible = true;
+                }, 1);
+            } else {
+                this.dialogVisible = false;
+                setTimeout(() => {
```
anri-asaturov: 1. save return value of this `setTimeout` in this.unmountTimeout.
2. Inside the timeout function put this.unmountTimeout = null
2. after line 43 put `if(this.unmountTimeout) clearTimeout(this.unmountTimeout)`

If not sure why - ask me.

suprko: Pushed an update - like that?

What's the difference between doing it like this and using `clearTimeout`?

suprko: oh wait. misread your comment. editing :sweat_smile:

src/ui/AutoUpdateDialog.jsx
```diff
@@ -2,7 +2,7 @@ const React = require('react');
 const { observable } = require('mobx');
 const { observer } = require('mobx-react');
 const { t } = require('peerio-translator');
-const { Dialog } = require('~/react-toolbox');
+const { Dialog } = require('~/peer-ui');
```
anri-asaturov: You can delete this entire file

suprko: OK! There are a few files that are not being used but I wasn't sure which may come back and which will not be needed again.

anri-asaturov: We can recover files from git history, clean that stuff up :-D

anri-asaturov: Tell me when to merge

suprko:
@anri-asaturov take another look at `Dialog.jsx`. Due to the way Dialogs are used in the app there are a lot of places where they are mounted but not rendered, so `keyup` subscription/unsub would not work properly in `component...Mount` functions and this was causing console errors. Had to move them into the `reaction` function.

If this looks OK then yes, ready to merge.

anri-asaturov:
@suprko add  `window.removeEventListener('keyup', this.handleEscKey);` to conponentWillUnmount and we are safe

anri-asaturov:
On another note - please refactor it a bit, make smaller functions out of the reaction so it's easier to read

suprko:
@anri-asaturov `componentWillUnmount` doesn't work, some Dialogs are always mounted, just not rendered/`active`. e.g. the ContactProfile popup when you click on an avatar

anri-asaturov:
@suprko I meant 'add another  window.removeEventListener('keyup', this.handleEscKey);' keep the one that's already in the reaction

suprko:
@anri-asaturov Ah I see :+1:

suprko:
@anri-asaturov done

dchest:
I noticed both old and new <Dialog> sometimes allow changing focus on the content that it covers by pressing Tab. 

tabbing


If I remember correctly, this can be prevented by:

1. Setting focus on the dialog element when it appears.
2. Catching onBlur event, checking if the focus is outside the container, and if it is, putting it back inside the dialog.

suprko:
@dchest good catch. I'll look into it

dchest:
@suprko turns out native `<dialog>` is already in Electron, perhaps it would be easier to use it? I check and it handles focus nicely.

https://keithjgrant.com/posts/2018/meet-the-new-dialog-element/

Not sure if backdrop is stylable, but the dialog itself is:

screen shot 2018-01-17 at 17 27 21

anri-asaturov:
great idea @dchest

PR #209, opened 2018-01-16
chore: replace react-toolbox <List> and related components

suprko:
https://app.clubhouse.io/peerio/story/1334/desktop-replace-list-components
 
#### Testing instructions  
Tested by me as we went along, component by component. I think it's safe to test on `dev`.

----
### Repository owner


src/ui/settings/components/ProfileSettings.jsx
```diff
@@ -133,39 +133,41 @@ class Profile extends React.Component {
...
+                                        className={css(
+                                            'email-row', { unconfirmed: !a.confirmed }
+                                        )}
+                                        caption={a.address}
+                                        legend={
+                                            a.confirmed
+                                                ? (a.primary ? t('title_primaryEmail') : null)
+                                                : t('error_unconfirmedEmail')
+                                        }
+                                        leftContent={
```
anri-asaturov: Let's maybe consider doing this via nested tags instead?

suprko: @anri-asaturov done

src/ui/settings/components/ProfileSettings.jsx
```diff
@@ -142,31 +142,38 @@ class Profile extends React.Component {
...
                                                     icon="radio_button_unchecked"
                                                     onClick={() => { this.makePrimary(a.address); }}
                                                     theme="small no-hover"
                                                 />
-                                        }
-                                        rightContent={
-                                            a.confirmed && !a.primary
+                                            }
+                                        </div>
+                                        <div className="main-content">
```
anri-asaturov: can we avoid this div?
make the whole thing a bit less verbose? maybe add stateless components?

suprko: As in, make a new `EmailRow` component or something like that?

anri-asaturov: As in think about the API usability of the software you write, how easy would it be for other devs to figure out how to use your List component?

I didn't think about it much but maybe smth like
```
<ListItem>
  <Caption>caption</Caption>
  <Icon position="left">blah</Icon>
   main content
</ListItem>
```

suprko: Wouldn't these tags clutter the imports though? Now every time you use a `ListItem` you'll also need to require `Caption`, `Icon`, `Legend`, or whatever else. I'm also not sure how `<Caption>` is more/less intuitive than `caption` as a prop, seems essentially the same...?

anri-asaturov: I'm not 100% against props, but prop that takes 10 lines of code is wrong probably, maybe then extract it so it's more readable

Nothing wrong with import clutter too, as long as i don't need to remember class names

src/ui/settings/components/ProfileSettings.jsx
```diff
@@ -133,39 +133,39 @@ class Profile extends React.Component {
...
+                                    : t('error_unconfirmedEmail');
+
+                                const rightButton = a.confirmed && !a.primary
+                                    ? (<Button tooltip={t('button_delete')} icon="delete"
+                                        onClick={() => { this.removeEmail(a.address); }}
+                                        theme="no-hover"
+                                    />)
+                                    : this.renderButton(a);
+
+                                return (
```
suprko: OK lmk what you think of this. Extracted out the props so it's more readable, structure of `ListItem` is still the same.

For better or worse this `ListItem` functions the same as RTB's... `leftContent` is the equivalent of `leftActions`. It may be possible down the road to split it into separate components or even something like `<List theme="simple">` but right now there's too much mix-and-match between `leftContent`, `caption`, `legend`, various custom content, etc... Something to revisit after design audit?

suprko: @anri-asaturov ^

PR #212, opened 2018-01-19
chore: replace RTB <Menu>

suprko:
 https://app.clubhouse.io/peerio/story/1342/desktop-replace-menu-components
https://app.clubhouse.io/peerio/story/2116/desktop-fix-pop-up-menu-margins-and-font-size
https://app.clubhouse.io/peerio/story/1368/desktop-menu-for-first-file-in-recent-file-list-cut-off

#### Testing instructions  
New `<Menu>` component! Replaced throughout app
* AppNav
* "Share files with this chat" plus button
* Three dots menu in
  * "Recent files" list in sidebar
  * Members list in sidebar
  * Files and folders list in Files view
* Dev tools
* Error message under failed message ("Send again" etc.)

----
### Repository owner


src/ui/AppNav.jsx
```diff
@@ -105,36 +105,74 @@ class AppNav extends React.Component {
...
+            {
+                value: 'help',
+                icon: 'help',
+                caption: 'Help',
+                onClick: this.toHelp
+            },
+            config.disablePayments || User.current.hasActivePlans
+                ? null : { divider: true },
+            config.disablePayments || User.current.hasActivePlans
+                ? null : {
```
suprko: I was unable to combine these two into a single condition and I don't really know why. `condition ? null : ({ k: val}, {k: val})` would only produce the 2nd object. I think there's something more fundamental about JS objects that I'm not getting...

Anyway this works with 2 separate bools...

suprko: Also when condition is false, you get a `null` object in the array... not ideal :expressionless:

src/ui/AppNav.jsx
```diff
@@ -105,36 +105,74 @@ class AppNav extends React.Component {
     render() {
         const contact = contactStore.getContact(User.current.username);
         const { primaryAddressConfirmed } = User.current;
+
+        const menuItems = [
```
suprko: In general what do you think of `menuItems` as object array, as opposed to using a `<MenuItem>` component? I originally liked it more because it seemed cleaner... but then we get the weird issue with conditional elements, and also the `{ divider }` object seems hacky

suprko:
@anri-asaturov @seavan see comments for my specific questions/concerns

src/ui/AppNav.jsx
suprko: The more I think about it the less it makes sense. Leaning towards rewriting this with `<MenuItem>` component

src/ui/AppNav.jsx
anri-asaturov: yes, you can't put it in one trenary in array initializer,
you can do smth like
```
if(condition){
items.push();
items.push()
}```

Comment (anri-asaturov) [reply src/ui/AppNav.jsx]: >condition ? null : ({ k: val}, {k: val}) would only produce the 2nd object.

js evaluates your ternary operator and returns this expression `({ k: val}, {k: val})` which is really 2 comma separated expressions, and the way js works - it will return the last one, so the last one will be used to insert an item to the array.

File: src/ui/AppNav.jsx (hunk 2 above)
Comment (anri-asaturov) [reply src/ui/AppNav.jsx]: Yes, MenuItem is good

**Comment by suprko** [comment]
@anri-asaturov @seavan switched to `MenuItem`, had to do some magic to take care of "unique key" console errors... Let me know what you think of overall component design though, as well as the issues mentioned in "Things I don't like" in PR comment

File: src/ui-library/MenuItem.jsx [hunk 3]
```diff
@@ -0,0 +1,57 @@
...
+    tooltipPosition string
+    ----------------------------------------
+*/
+
+class MenuItem extends React.Component {
+    render() {
+        const { value, icon, customIcon, caption, onClick, className } = this.props;
+        return (
+            <div
+                key={this.props.key || 'menu-item'}
```
anri-asaturov: this is not needed

anri-asaturov: you should set key when you create MenuItem, where it actually has siblings.

suprko: :thinking: yeah already done here https://github.com/PeerioTechnologies/peerio-desktop/blob/ea9c88df6c6dcc12230fce2b934723d7db557fb2/src/ui-library/MenuGlobal.jsx#L32

For some reason I thought it wasn't working without the dummy value but it must have been something else that was still throwing the errors. Fixed

src/ui-library/MenuGlobal.jsx
```diff
@@ -0,0 +1,64 @@
...
+                this.menuItems.push(item);
+            }
+        });
+        this.menuVisible = true;
+    }
+
+    @action.bound setInactive() {
+        this.menuVisible = false;
+
+        this.unmountTimeout = setTimeout(() => {
```
suprko: When I removed the timeout, menu didn't work at all, and I realised that the menu was disappearing instantly and therefore not registering the `MenuItem` click. Adding 1 frame `setTimeout` fixes it but it feels wrong. Gonna keep thinking through how to make it work in a better way...

src/ui/files/Files.jsx
```diff
@@ -49,10 +49,16 @@ class Files extends React.Component {
         fileStore.filterByName(val);
     };
 
+    // Establish a "selected folder" on FolderActions click, for use with FolderAction functions
+    @observable folderMenuSelected;
+    @action.bound folderMenuSelect(ev) {
+        this.folderMenuSelected = getFolderByEvent(ev);
```
suprko: Various `FolderActions` functions target folder by using `getFolderEvent` helper function, which recursively searches through parent chain for `data-folderid` attr. The problem is, new `MenuGlobal` component means that the click is registered outside of this chain (`MenuGlobal` is portalled to app root). So this function sets `folderMenuSelected` when any `Menu` inside a `FolderActions` is clicked.

src/ui/files/Files.jsx
```diff
@@ -49,10 +49,16 @@ class Files extends React.Component {
...
+    }
+
     @observable moveFolderVisible = false;
 
     @action.bound moveFolder(ev) {
-        const folder = getFolderByEvent(ev);
+        const folder = getFolderByEvent(ev) || this.folderMenuSelected;
```
suprko: (continuing from line 55 comment)

The problem is that most FolderActions functions are also shared by Breadcrumb. So we need to account for both possibilities -- set folder by click event target, or if that's null, use `folderMenuSelected`. Alternative is to separate each function into two distinct functions.

suprko:
@anri-asaturov @seavan <Menu> has been replaced throughout the app. Still a few things that bug me, see comments on individual files.

suprko:
@anri-asaturov @seavan more changes, at this point it is mostly complete aside from the remaining thing in "Things I don'd like" in main PR comment.

The `ev.target` problem in Files actually came up several other places, so there is some restructuring of how MenuItem click events work in InlineFiles, MemberSection, FilesSection...

Also not sure if I handled the close-menu-on-scroll behaviour correctly, see individual comment on `Menu.jsx`

src/ui-library/Menu.jsx
```diff
@@ -0,0 +1,164 @@
...
+
+@observer
+class Menu extends React.Component {
+    scrollContainer;
+    menuButtonRef;
+    @observable menuActive;
+
+    @action.bound setRef(ref) {
+        this.menuButtonRef = ref;
+        this.scrollContainer = getParentWithAttribute(ref, 'data-scrollable');
```
suprko: Scroll event listener needs to listen on a specific element, so I created `getParentWithAttribute` helper function (based on `getAttributeInParent`) that looks for `data-scrollable` prop, and added `data-scrollable` prop to the elements that have `overflow-y:auto`. Not sure if this is the best way to find closest parent that scrolls.

src/ui-library/Menu.jsx
```diff
@@ -0,0 +1,164 @@
...
+
+    @action.bound handleClick(ev) {
+        if (!elementHasClass(ev.target, 'p-menu-item')) {
+            this.hideMenu();
+        } else {
+            // If clicking on a <MenuItem>, reset *this* menu, but do not hide <MenuGlobal>
+            // It will be handled by <MenuItem>'s onClick event
+            this.menuActive = false;
+            this.clearListeners();
+        }
```
suprko: This is the part I feel most uneasy about. Previously, `MenuGlobal` onClick had a timeout of 1, so that hiding the menu would fire *after* the `MenuItem` click registered.

To solve this, I've added a condition to this `handleClick` function so that, if click target is `MenuItem`, *this* `Menu` component is closed, but the `MenuGlobal` is left open, and closing it is left up to the `MenuItem` component itself.

```
+    handleClick = (ev) => {
+        this.props.onClick(ev);
+        uiStore.clearGlobalMenu();
+    }
```
It works, but it feels weird because we're splitting up the "close the menu" action across two components... So I don't know if this is better or worse than the `setTimeout` option

src/ui/files/components/FileActions.jsx
```diff
@@ -13,47 +13,72 @@ function FileActions(props) {
...
-                    icon="delete"
-                    onClick={props.onDelete} />}
-            </IconMenu>
-        </div>
-    );
+
+    render() {
+        return (
+            <div className="item-actions" onClick={this.props.onRowClick}>
+                <Menu icon="more_vert" position="bottom-right" onClick={this.props.onMenuClick}>
```
seavan: you can, and should, provide data-folderid here, @suprko 
I don't like setting separate folderMenuSelected etc in the higher up component, it feels spagetti-like.
if you set data-folderid of any parent container, including menu itself, getFolderByEvent will work.

suprko: `data-folderid` on `Menu` won't work. In the DOM, `Menu` is just the button; `MenuItem`s get passed to `MenuGlobal` component, which is 

I could put `data-folderid` on each individual `MenuItem` though, is that still better than `folderSelected`?

suprko: Or pass `Menu` data props to `MenuGlobal`?

src/ui-library/Menu.jsx
```diff
@@ -0,0 +1,164 @@
...
+
+    @action.bound menuClick() {
+        uiStore.clearGlobalMenu();
+
+        // Need to deal with this annoying thing: 1 child = object, >1 children = array
+        if (this.props.children.length > 1) {
+            uiStore.globalMenuContent.items = this.props.children;
+        } else {
+            uiStore.globalMenuContent.items = [this.props.children];
+        }
```
seavan: you could just do uiStore.globalMenuContent = this.props for example
and move the logic of rendering this.props.children.length > 1 to GlobalMenu

then, in the MenuGlobal you could do <div myProperty1="something" {...uiStore.globalMenuContent}>{uiStore.globalMenuContent.children}</div>

or something

suprko: Passing all `props` to `globalMenuContent` is unwise I think, for example Menu's `onClick` will get passed to MenuGlobal too...

I'm trying a thing right now, basically going to *just* pass data props, will update you when it's done.

seavan: you can override anything you want:

<div {...someProps} onClick={myClick}>something</div>

it's a normal pattern in react - passing all the props and overriding them as needed,

seavan: @suprko fixed

suprko:
@seavan OK fixed. bigger task than I thought, I had to refactor MembersSection menus

seavan:
@suprko is it still WIP or do you want a review?

suprko:
@seavan pretty sure this is done and ready for review but we're on code freeze until release, right?

src/ui-library/Menu.jsx
seavan: setTimeout is better in my opinion.
it could also be just setTimeout(..., 0) or just setTimeout(...). should work too.

src/ui-library/Menu.jsx
```diff
@@ -0,0 +1,158 @@
...
+                    left: `${left}px`
+                };
+                break;
+            case 'bottom-right':
+                style = {
+                    bottom: `${windowY - top - height}px`,
+                    right: `${windowX - left - width}px`
+                };
+                break;
+        }
```
seavan: why recreate object each time?
you can do:
const style = {};
const [ pos1, pos2 ] = position.split('-');
if (pos1 === 'top') style.top = `${top}px` else style.bottom = `${windowY - top - height}px`;
if (pos2 === 'left') style.left = `${left}px` else style.right = `${windowX - left - width}px`;

src/ui/files/components/FileActions.jsx
```diff
@@ -13,47 +13,76 @@ function FileActions(props) {
...
                 {/* onClick={noop} */}
-                {/* className={css({ disabled: props.newFolderDisabled })} /> */}
-                {props.deleteable && <MenuDivider />}
-                {props.deleteable && <MenuItem caption={t('button_delete')}
-                    icon="delete"
-                    onClick={props.onDelete} />}
-            </IconMenu>
-        </div>
-    );
+                {/* className={css({ disabled: this.props.newFolderDisabled })} /> */}
```
seavan: remove commented code if it's not a WIP anymore

src/ui/files/components/FileActions.jsx
```diff
@@ -13,47 +13,76 @@ function FileActions(props) {
...
-                    className={css('reverse-icon', { disabled: props.shareDisabled })} />
+
+    render() {
+        return (
+            <Menu
+                className="item-actions"
+                icon="more_vert"
+                position="bottom-right"
+                onClick={this.props.onMenuClick}
+                {...getDataProps(this.props)}
```
seavan: I kinda don't like scanning props each time for data attributes, still would've preferred plain {...this.props} in the beginning of att-list. @anri-asaturov what do you think

src/ui/files/components/FileActions.jsx
```diff
@@ -13,47 +13,76 @@ function FileActions(props) {
...
+                        this.props.downloadDisabled
+                        ? <ProgressBar
+                            key="0"
+                            type="linear"
+                            mode="indeterminate"
+                            multicolor
+                            className="processing-file"
+                        />
+                        : null
+                    */}
```
seavan: remove commented code if it's not a WIP anymore

seavan:
@anri-asaturov take a look please.

src/ui-library/Menu.jsx
suprko: Fixed the other things, will get to this one tomorrow morning

src/ui/files/components/FileActions.jsx
anri-asaturov: I thought there was some problem doing that, @suprko ?

src/ui-library/Menu.jsx
anri-asaturov: It's ok for now, if someone forgets it - the worst part is the menu will not scroll right?
And we can refactor it later if we have better ideas

src/stores/ui-store.js
```diff
@@ -33,6 +33,20 @@ class UIStore {
         }
     }
 
+    // Content for <Menu>s gets passed to <GlobalMenu> via globalMenuContent
+    // contains array of menuItems & style object for positioning
+    @observable globalMenuContent = {
+        props: {},
+        style: {}
+    };
```
anri-asaturov: You couldn't figure out portals or what happened? :-)

suprko: I thought you told me to use ui-store??? to pass props from individual menu items to a global menu component. Portal renders a thing outside of the parent component, but we'd still have separate menu element for each `<Menu>` that's in the app, as opposed to a single global component.

I mean tbh I have no problem making menus the same way we made tooltip if that's preferred.

src/ui/files/components/FileActions.jsx
suprko: Not really a problem per se. FileActions has (more than?) a dozen props and they're used later in the component, so if we pass all of those props to Menu then Menu ends up with random things like `shareable` and `onShare` props. They won't affect anything but if it's better to pass it all props instead of using `getDataProps` it will still work the same.

src/stores/ui-store.js
anri-asaturov: Yes, if there's no big issues associated with that, it's better to have decoupled component. Performance wise it will be the same right? Same amount of renders.

suprko: re: performance, I'm not sure... Currently the `MenuItem`s don't actually show in DOM until the `Menu` button is clicked, and then all `MenuItem`s are sent to `MenuGlobal` and rendered there. Is that saving renders, or just pushing them around?

anri-asaturov: Well you'll render MenuItems in a portal, which also won't show(render) until you click menu button

suprko: Ah, yeah! Like Dialog. I was thinking of Tooltip, which is always in DOM. OK 👍

src/ui/files/Files.jsx
seavan: @suprko better to set correct data-folderid in breadcrumb also and pass it to menu.

src/ui/files/Files.jsx
seavan: as discussed, @suprko, data-folderid should be passed to MenuGlobal so that we don't have to introduce this intermediate variables.

suprko: @seavan sorry, missed it. done

anri-asaturov:
@suprko reviewing? merging? is this ready?

suprko:
@anri-asaturov reviewing I thought? I requested reviews on Friday...

src/ui/files/components/FileActions.jsx
anri-asaturov: It's safer, yes

anri-asaturov:
@suprko merging or you want separate test or maybe to add something else?

suprko:
@anri-asaturov I think merge and we can test on dev like we did with other components. I've got another branch already for the next set of components

PR #213, opened 2018-01-20
fix: use observer

seavan:

don't forget to put observer @suprko 

https://app.clubhouse.io/peerio/story/2496/desktop-doesn-t-update-avatar


anri-asaturov:
What's the motivation behind converting all basic component to be observer?

seavan:
because the component won't update if it's not an observer on props update, otherwise.

everything that uses props should be an observable.

anri-asaturov:
Yeah, i know that, for things like Avatar.

Basic components like Button or ListItem - I have my doubts. Usually you don't change Button props without re-rendering parent component, so it will update naturally.  React-toolbox components were not observers.

Is there a bug or smth related to this changes (except Avatar)?

seavan:
it'll update naturally only if the parent component is an observer. if the parent component is not, it won't, even if you pass observables to them.

but i think you have a point, we may remove it for basic components for performance issues.

seavan:
left it only for Avatar, @anri-asaturov

PR #220, opened 2018-01-25
feat: no auto adding contacts on contact search

seavan:
 
https://app.clubhouse.io/peerio/story/2493/desktop-do-not-add-contacts-without-explicit-action

https://github.com/PeerioTechnologies/peerio-icebear/pull/130


src/ui/chat/NewChat.jsx
```diff
@@ -16,9 +16,8 @@ class NewChat extends React.Component {
         chatStore.deactivateCurrentChat();
     }
 
-    handleAccept = async (selected) => {
+    handleAccept = (selected) => {
         this.waiting = true;
-        await Contact.ensureAllLoaded(selected);
```
anri-asaturov: This was supposed to wait until contacts are loaded, maybe user made a mistake, does user picker handle that properly? (will contacts that have `loading===true` never end up in `selected` or will accepting user picker selection be not possible before all contacts are loaded?)

seavan: picker doesn't allow for contacts unless they are loaded and there's also same check in icebear (ensureAllLoaded)

PR #221, opened 2018-01-26
fix: set default peerioContentEnabled to false

suprko:
https://app.clubhouse.io/peerio/story/2721/desktop-unable-to-disable-image-previews
 
#### Testing instructions  
Images should not load when "Not now" or "Disabled" selected from Peerio content consent, should load when "Enabled" selected.

----
### Repository owner


suprko:
The cause of this bug was that `peerioContentEnabled` was set to default true in ui-store, so `_fileReaction` would call `tryToCacheTemporarily` as soon as file showed up, before user had answered consent. I don't know if there was a reason for `peerioContentEnabled` default to be set to true initially, but setting it to false seems to fix this and doesn't affect the rest of the consent/preview flow.

seavan:
@suprko inline images from peerio should be enabled by default, this is why peerioContentEnabled is set to true.

external images, on the other hand, should have confirmation dialog "do you want to enable blahblah". for that there's a separate flag.

I don't think setting peerioContentEnabled to false is correct decision. this way peerio inline images would be disabled until you send an external link.

suprko:
@seavan I don't really understand then, what's the purpose of having the Peerio image consent form if it's enabled by default?

screenshot from 2018-01-26 09 42 47

suprko:
> I don't think setting peerioContentEnabled to false is correct decision. this way peerio inline images would be disabled until you send an external link.

No, Peerio inline images would be disabled until someone uploads an image to your chat, and then you'll be asked for consent. They're separate forms.

seavan:
consent is for external images, not for peerio ones. we have two types. external images are the one which go by the link.

suprko:
We have consent for both. `peerioContentConsented` is for Peerio images, `externalContentConsented` is for URL images

seavan:
image

peerioContentConsented seems to be unneeded. it's not used in mobile.

i'll confirm the logic on meeting.

suprko:
I'm just fixing the bug I was given lol. If we want to remove Peerio image preview consent form, then that has to go back to design, no?

seavan:
yeah I need to confirm it. behaviour on mobile is 100% that inline images don't have consent for them. and I've thought the bug was about external images, not internal.

suprko:
Eren confirms this is for uploaded/Peerio images https://app.clubhouse.io/peerio/story/2721/desktop-unable-to-disable-image-previews

PR #222, opened 2018-01-29
feat: don't allow more than one instance of app

dchest:
In production, if another instance is launched, it will detect that the
app is already running, and quit. The first instance will bring itself
to the front.

To turn this off, pass `--allow-multiple-instances` command-line argument
(on macOS, the app needs to be launched from command-line for that).

^ this flag is for @ebilg :)

#### Testing instructions  

Testing needs to be performed with production build (`npm run internal-build`) on all operating systems.

1. Launch the app.
2. Launch it again.
3. The second instance should quit before it appears and the first one should re-appear and become focused.
4. Minimize the app.
5. Launch again.
6. The second instance should quit before it appears and the first one should unminimize itself and become focused.

Please ensure that the first instance does bring itself to the front on Windows, I'm a bit concerned about it.

If you want to test `--allow-multiple-instances` note that it needs to be passed to the very first instance and every additional instances to launch.

----
### Repository owner


src/main.js
```diff
@@ -4,14 +4,36 @@ if (process.env.NODE_ENV !== 'development') {
...
+        if (mainWindow) {
+            if (mainWindow.isMinimized()) {
+                mainWindow.restore();
+            }
+            mainWindow.focus();
+        }
+    });
+    if (isAnotherInstance) {
+        console.log('Another instance is already running, quitting.');
+        app.quit();
```
anri-asaturov: This doesn't exit immediately, file executes till the end and window shows briefly. Use `process.exit()`.

dchest: Is `process.exit()` safe to use? Maybe I should just return from here?

anri-asaturov: `return` in global scope is illegal, i tested process.exit() - it works

dchest: Done

anri-asaturov:
Ready to test. @ebilg consider the option to allow running another instance if it was started from another folder, it's possible to do.

dchest:
> I wonder how this will affect restart upon app update.

Right, need to check the effect on updaters. 

* On Linux restart doesn't work anyway. 
* On Windows it launches setup.exe and quits... however if it takes a lot to quit (deleting caches), it may prevent the updated version from launching.
* On macOS -- not sure, but I believe it launches an external process to update, so same thing applies.

Hold on, will figure out how to do it, probably just disable this when installing update.


> Also, do we maybe want to allow multiple instances if they have different working directories?

dchest:
> Also, do we maybe want to allow multiple instances if they have different working directories?

Why?

```
$ cd ~/Desktop && /Applications/Peerio 2/.../Peerio 2
$ cd /Applications && /Applications/Peerio 2/.../Peerio 2
```

I don't think it should be any different.

anri-asaturov:
I was talking more about regular users who might have app copes in different folders? Maybe even different versions... Can't say for sure why, just asking to think about it, there are some weird cases.

dchest:
Ah, I see. I think many copies still shouldn't launch more instances. Firefox, for example, doesn't allow this.

dchest:
Looks like it depends on DIR_USER_DATA, not app id: https://github.com/electron/electron/blob/c49cb29ddf3368daf279bd60c007f9c015bc834c/atom/browser/api/atom_api_app.cc#L863

ebilg:
looks good

PR #223, opened 2018-01-29
chore: replace rtb ProgressBar, Chip, Tabs, Switch

suprko:
https://app.clubhouse.io/peerio/story/1341/desktop-replace-chip-components
https://app.clubhouse.io/peerio/story/1339/desktop-replace-tabs-components
https://app.clubhouse.io/peerio/story/1338/desktop-replace-switch-component
https://app.clubhouse.io/peerio/story/1335/desktop-replace-progressbar-component

#### Testing instructions  
similar to other RTB replacements, I was testing as I went along but may have missed something, best strategy is probably to test on `dev`?

----
### Repository owner


src/ui/settings/components/Preferences.jsx
```diff
@@ -11,42 +10,42 @@ const { fileStore } = require('peerio-icebear');
...
-        uiStore.prefs.errorSoundsEnabled = value;
+    onErrorSoundsChanged(ev) {
+        uiStore.prefs.errorSoundsEnabled = ev.target.checked;
     }
 
-    onMentionSoundsChanged(value) {
+    onMentionSoundsChanged(ev) {
         uiStore.prefs.messageSoundsEnabled = false;
-        uiStore.prefs.mentionSoundsEnabled = value;
+        uiStore.prefs.mentionSoundsEnabled = ev.target.checked;
```
anri-asaturov: Duuuude, I see you're not lazy. Could've just maintained contract (pass bool instead of event)

suprko: Just want to clarify how this would work: would it be in the Switch component, make the `onChange` function `this.props.onChange(this.props.checked)`?
https://github.com/PeerioTechnologies/peerio-desktop/blob/9fb70363a6afc7803585845fbdc9ac478559650b/src/ui-library/Switch.jsx#L31

anri-asaturov: yes

suprko: Worth fixing? I always feel weird using click event targets...

anri-asaturov: Up to you.

PR #225, opened 2018-01-31
 feat: don't quit app when closing window on macOS

dchest:
Notes:

There's a new menu item on macOS in Window menu to show the main window. Ideally, it would work like the one that automatically appears at the bottom of the menu ("Peerio 2 DEV" on screenshot) when the window is shown, but will stay visible even if the window is closed ("Peerio 2 DEV" disappears), like in Calendar app:

screen shot 2018-01-31 at 13 31 38

screen shot 2018-01-31 at 13 33 53

However, Electron doesn't seem to support setting `setExcludedFromWindowsMenu:YES` on the window, so we can't hide the one that appears and disappears. 

Example from Mail app ("Message Viewer"):

screen shot 2018-01-31 at 13 39 30

For this reason I decided to call it "Main Window" (like in Twitter app) instead of `app.getName()` to avoid having two menu items with the same name when the window is shown.

The shortcut — `Cmd+0` is standard for all Apple's apps. If we'll have multiple accounts support in one window, then we can remove this item and instead use `Cmd+1`, `Cmd+2`... for accounts. If we don't, we can use these shortcuts for switching between chats, files, etc. in the sidebar. Then we can also remove "Main Window" item and put "Chats ⌘1", "Files ⌘2", etc there.

https://app.clubhouse.io/peerio/story/2562/desktop-do-not-quit-peerio-when-window-is-closed

#### Testing instructions  

On Mac:

1. Open Peerio.
2. Close the window.
3. Check that Peerio still runs.
4. Click on Peerio icon.
5. Check that window appears.
6. Close the window again.
7. In top menu, click Window > Main Window.
8. Check that window appears.
9. Quit the app (Cmd+Q or Peerio 2 > Quit Peerio 2).
10. Make sure the app did quit.

On Windows and Linux:

Just make sure that closing the window quits the app. No changes here.

----
### Repository owner


ebilg:
this is so cool!! can be merged

PR #230, opened 2018-02-02
style: 'person' icon in Rooms header + update avatar in ProfileSettings

suprko:
https://app.clubhouse.io/peerio/story/888/desktop-screen-room-user-icon
https://app.clubhouse.io/peerio/story/1550/desktop-improve-error-handling-when-updating-avatar

#### Testing instructions  
The main visible change: In room, the "person" icon is no longer a button that opens the sidebar. Instead is just an icon, mouseover shows a tooltip that lists all members of the room.

PR includes a small update to `<Avatar>` component for `size="full"`, and using `<Avatar>` in ProfileSettings view. This allows the ProfileSettings avatar image to retry if download fails. (Before it was a `background-image`, and those don't retry.)

It is still possible to break avatar images by uploading a large image and then breaking internet connection a lot during the download. A broken image will be cached by Electron and will be persistent after restart. However, IMO this problem is out of scope for this particular story, and has more to do with image retrying & caching in general (maybe even at the Electron level?), not specific to avatars.

----
### Repository owner


ebilg:
can be merged

src/ui/chat/ChatView.jsx
```diff
@@ -126,6 +126,10 @@ class ChatView extends React.Component {
     // assumes active chat exists, don't render if it doesn't
     renderHeader() {
         const chat = chatStore.activeChat;
+        const listMembers = chat.allParticipants.map(p => {
```
anri-asaturov: use `chat.participanUsernames`

suprko: `participantUsernames` gets usernames from `otherParticipants`, so self isn't included

anri-asaturov: Isn't this a tooltip? why do we want ourselves in a tooltip? @paulrdukes ?

anri-asaturov: In any case you can change participantUsernames in the sdk, bcs it's not used anywhere where it could break if you add current user. Also I see that mobile never uses it, maybe it does the same mapping somewhere would be cool to find it. @Karim94 ?

suprko: It looks kind of weird if the number says there are 2 members in chat, but the tooltip only lists 1 member, for example. Including self in the member list is also consistent with the room header: 
screenshot from 2018-02-05 08 47 31

suprko:
@anri-asaturov previously forgot about limiting tooltip member list to < 100 characters, that's been added now. please re-review

anri-asaturov:
@suprko any chance to limit string length with css?

anri-asaturov:
Cuz then you could just do `praticipantUsernames.join(', ')`

suprko:
@anri-asaturov tl;dr not really... `text-overflow:ellipsis` only works on single lines of text, multiline not possible.

The other thing is that we don't want individual usernames to get cut off. If adding a username makes the whole string over the limit, then don't show the username at all. We want to avoid for example "lucas, anri, eren, sa..."

anri-asaturov:
Oh boy. Maybe 'anri, lucas +10' then?

anri-asaturov:
Ready to test

ebilg:
can be merged

PR #235, opened 2018-02-06
 feat: ask to copy to Applications when running from DMG on macOS

dchest:
Please proofread the dialog messages!

https://app.clubhouse.io/peerio/story/2704/mac-detect-when-app-is-launched-from-dmg 

#### Testing instructions  

MacOS only.

#### A

1. Make internal macOS build `npm run internal-build-mac`.
2. (Backup prod version that's in `/Applications` if you have it to get back to it)
3. Open `dist/peerioXXX.dmg`.
4. Double-click Peerio in the mounted DMG to launch the app.
5. Check that you're asked to copy the app.
6. Click "Copy to Applications Folder".
7. Check that it copies and restarts (if there was Peerio 2 in the Applications folder, it will be put into Trash). Make sure the restarted app is the one from Applications (right-click on the icon in Dock > Options > Show in Finder).
8. Quit the app, then launch it from Applications folder manually to ensure that the dialog won't appear.
9. Eject the disk image. Make sure it ejects.

#### B

1. Repeats 4-5 from A.
2. Click "Do Not Copy".
3. Check that the app just launches.
4. Check that there's no "Checking for update..." messages in console (autoupdate is disabled when running from DMG).

#### C

1. Repeat B, but before step 2 click "Do not show this message again".
2. Launch the app from DMG again and check that it no longer shows the dialog and just launches from DMG.

*"Do not show this message again" state is stored in `~/Library/Application Support/Peerio 2/$system$_tinydb.json` under key `doNotCopyToApplicationsFolder`. Remove it to get the dialog back.*

### D

1. Launch already existing app from /Applications.
2. Launch the one from DMG and click "Copy..."
3. The launched copy should quit and the already running instance will be brought to front.

#### E

1. Upload DMG somewhere on the web (Dropbox/Google Drive/whatever)
2. Download it (into ~/Downloads, preferably using Safari)
3. Double click to launch.
4. Follow test case C, making sure that after copying application appears in /Applications folder. After applications restarts, it shouldn't show "File downloaded from the internet..." dialog.

### Under a Standard account

screen shot 2018-02-06 at 18 03 15

Create a standard account if there's none, and copy DMG somewhere to it has access to, e.g. 

Perform the same steps as in (A) for administrator account, but after clicking "Copy..." it should ask for administrator username and password. Enter them and check if Peerio was copied.

----
### Repository owner


ebilg:
Use the app name "Peerio 2" instead of the word "Application" for   'Application is running from a disk image' and 'It is not recommended to run application directly from the disk image. '

dchest:
More test cases:

### D

1. Upload DMG somewhere on the web.
2. Download it (into `~/Downloads`)
3. Double click to launch.
4. Follow test case C, making sure that after copying application appears in /Applications folder.

### E

1. Launch already existing app from /Applications.
2. Launch the one from DMG and click "Copy..."
3. ... *What happens next is undecided. If we overwrite the copy, the running one may become broken. We need detect the running copy, and either quit it, or complain that we can't copy.*

dchest:
@ebilg the coping code will be replaced with native Electron's which will be available once we update it after the current release, so if you haven't tested this PR yet, don't (if you've already tested, it needs to be retested after I replace the code 😬).

anri-asaturov:
Let's merge this in the next release any way

dchest:
Updated and ready for testing (test cases are also updated in the description).

Note that I couldn't get copying working from non-administrator account with new Electron's method. It always shows me the error after it asks for username/password of administrator, but I suspect this is because I had some changes to root account on my Mac. If it won't work for others, we can partically revert back to showing "Please copy the app to Applications" dialog for non-admin accounts as before and file a bug in Electron.

ebilg:
Issues:

1. For test case D running Peerio from the DMG foregrounds the currently running copy of Peerio

2. Test case E - I'm not sure what happened but I was not prompted to copy the app to Applications folder when running from DMG which was downloaded from the internet. I would get the dialog asking if I want to run this app which is downloaded from the internet, and then Peerio would just run.

3. Under a standard account copying to Applications folder fails after entering Admin credentials, error shows.

dchest:
Thanks!

1. Yes, this is expected — sorry if the test description wasn't clear. I made a separate story for that https://app.clubhouse.io/peerio/story/3624/mac-inform-users-that-they-should-close-another-copy-if-it-s-running-before-moving-to-applications-folder-from-dmg

2. Interesting. Will investigate and make changes to this PR.

3. Right, so not my computer than, bug in Electron. I'll revert to the previous "Please copy the app..." message for standard accounts then.

ebilg:
can be merged

PR #240, opened 2018-02-08
feat: room invite redesign [ch273]

seavan:
https://app.clubhouse.io/peerio/story/273/desktop-room-invites-redesigned-flow
https://github.com/PeerioTechnologies/peerio-icebear/pull/141

#### Testing instructions  
This is what works so far:
* Room invites are listed along with chats, in alphabetical order, according to mockups
* Clicking room invite opens `ChannelInvite` component
* Host avatar is shown
* Accepting and rejecting invite works with existing SDK
  * very minor modifications to SDK so far; just made channel activate after accepting invite, which is how it used to work

SDK fixes (future):
* Need to be able to mark invites as `active` in SDK, the way rooms/DMs can be marked as `active`
  * Selecting a room invite does not currently set its entry in `ChatList` to `active` style, i.e. white text on blue background
  * ~~After declining an invite, selecting another room invite still shows the "invite declined" screen, because all room invites share the same component (will fix on client side for now)~~
* Similarly, need to be able to mark invites as `read`/`unread` in SDK
  * Selecting a room invite does not currently remove the `unread` style, i.e. semibold text
* No list of participants other than host

----
### Repository owner


src/stores/ui-store.js
```diff
@@ -58,10 +58,13 @@ class UIStore {
...
     observePreference(key, dbName, localStore) {
         const prefKey = `pref_${key}`;
-        TinyDb[dbName].getValue(prefKey)
+        return TinyDb[dbName].getValue(prefKey)
             .then((loadedValue) => {
                 if (loadedValue || loadedValue === false) {
                     localStore[key] = loadedValue;
+                } else {
+                    // TODO: to be removed in next release after 2.108
+                    localStore[key] = null;
```
anri-asaturov: So, for example, we convert value `0` to `null` and pray nothing breaks?

anri-asaturov: Point is: why the carpet bombing? You know which keys are problematic - fix just them

anri-asaturov: Ok, nvm, all values are boolean here.

anri-asaturov: But this sets all props to null for new accounts, right? I don't understand how this is ok?

anri-asaturov: Ok. confirmed, this is a bug, Eren will create one in CH

anri-asaturov: This is a bug: it resets all the default values for the accounts loggin in for the first time on the device. Do this reset only for the property in question

src/ui/chat/components/ChannelInvite.jsx
```diff
@@ -0,0 +1,106 @@
...
+                <div className="invite-content decline-content">
+                    <div className="emoji-double emojione-32-diversity _270c-1f3fc" alt="✌️" title=":v:" />
+                    <div className="text">
+                        {t('title_userOut', { name: User.current.username })}
+                    </div>
+                </div>
+            </div>
+        );
+
+        return ReactDOM.createPortal(
```
anri-asaturov: What is this? 😨

suprko: A portal. Because we need ChannelInvite to render in `.messages` div, not in ChatList.

anri-asaturov: Why not put <ChannelInvite> to the messages div?

suprko: So then, store active room invite as an object in `ui-store`?

src/ui/chat/components/ChatList.jsx
```diff
@@ -77,35 +132,46 @@ class ChatList extends React.Component {
...
+                                                    className={
+                                                        css('room-item',
+                                                            {
+                                                                active: r.active,
+                                                                unread: r.unreadCount > 0
+                                                            }
+                                                        )
+                                                    }
+                                                    caption={`# ${r.name}`}
+                                                    onClick={() => this.activateChat(r)}
```
anri-asaturov: Make it without arrow while you're at it please

src/ui/chat/components/ChatList.jsx
```diff
@@ -77,35 +132,46 @@ class ChatList extends React.Component {
...
+                                                    rightContent={
+                                                        ((!r.active || r.newMessagesMarkerPos) && r.unreadCount > 0)
+                                                            ? this.getNotificationIcon(r)
+                                                            : null
+                                                    } />
+                                            )
+                                            : (
+                                                <ListItem
+                                                    key={`${r.kegDbId}${r.username}${r.timestamp}`}
```
anri-asaturov: why the composite key?

src/ui/chat/components/ChatList.jsx
```diff
@@ -54,8 +79,46 @@ class ChatList extends React.Component {
         );
     }
 
+    @observable channelInviteActive;
+    @observable selectedInvite;
+
+    @action.bound activateInvite(ev) {
+        chatStore.deactivateCurrentChat();
+
+        const invite = getAttributeInParentChain(ev.target, 'data-invite').split(',');
```
suprko: @anri-asaturov @seavan this seems really hacky but I couldn't think of a better way to pass these props to ChannelInvite as an object...? Didn't realize until now that `data` props can only be strings, I think?

anri-asaturov: Channel invite can be identified by kegDbId, no need for composite id.

suprko: Then we'd need to get the `username` and `channelName` by iterating through `chatInviteStore.received` array looking for a matching value for this `kegDbId`. Isn't it more performant to just pass the strings directly from the clicked `ListItem`?

anri-asaturov: Iterating is fine. id is id, it's not supposed to pass data.

anri-asaturov: Invites won't ever be so large to make a difference.

src/ui/chat/components/ChatList.jsx
```diff
@@ -54,8 +79,45 @@ class ChatList extends React.Component {
...
     }
 
+    @observable channelInviteActive;
+    @observable selectedInvite;
+
+    @action.bound activateInvite(ev) {
+        chatStore.deactivateCurrentChat();
+
+        const kegDbId = getAttributeInParentChain(ev.target, 'data-kegdbid');
+        const { channelName, username } = chatInviteStore.received.find(obj => {
```
suprko: @anri-asaturov fixed, very fancy now ✨

anri-asaturov: What is for whatever reason (invite got deleted) `.find` returns null?

suprko: Not sure 🤷‍ is that possible? If invite is deleted then ListItem disappears. I suppose it's possible for invite to be deleted at the exact moment that the user clicks the ListItem... I don't know what the intended behaviour would be in this case though

anri-asaturov: It is possible. You should check if something was found. If not - just return. Current 'fancy' code will throw

src/ui/chat/components/ChannelInvite.jsx
anri-asaturov: I guess for now yes. Don't forget.

src/ui/chat/components/ChatList.jsx
```diff
@@ -54,8 +79,52 @@ class ChatList extends React.Component {
...
+
+    @action.bound activateInvite(ev) {
+        chatStore.deactivateCurrentChat();
+
+        const kegDbId = getAttributeInParentChain(ev.target, 'data-kegdbid');
+        const invite = chatInviteStore.received.find(obj => {
+            return obj.kegDbId === kegDbId;
+        });
+
+        if (!invite) return;
```
suprko: @anri-asaturov ^

src/ui/chat/components/ChannelInvite.jsx
anri-asaturov: This is not a place for portal, 
```
Portals provide a first-class way to render children into a DOM node that exists outside the DOM hierarchy of the parent component.```

File: src/ui/contact/NewContact.jsx [hunk 8]
```diff
@@ -66,6 +66,7 @@ class NewContact extends React.Component {
                 }
             } else {
                 this.query = '';
+                warnings.add(t('title_contactAdded', { name: c.fullNameAndUsername }));
```
anri-asaturov: This is a bug. In previous versions this code was calling `contactStore.addContact()` which was making the contact favorite. Now we're just creating tofu for it. @ebilg was there some requirements change in here?

anri-asaturov: Ok, Eren says this is by new design.

PR #243, opened 2018-02-09
chore: change circular ProgressBar to SVG

suprko:
https://app.clubhouse.io/peerio/story/3052/desktop-switch-progressbar-to-svg

#### Testing instructions  
I was checking as I went, but maybe could use more eyes on it? We're checking that `<ProgressBar>` looks alright wherever it's used with `type="circular"`. It may look slightly different in some places, but it shouldn't look outright wrong (e.g. large spinner where it should be a small spinner). We have permission from Paul to deviate a bit from previous design because we'll be replacing this soon anyway with upcoming rebrand.

Circular progress bar/spinner shows up in:
* ChatView, when clicking the pin button to pin/unpin a chat
* MessageList
  * in the middle when messages are loading
  * at top and bottom when you scroll through messages
* After creating a new room
* KegEditor... somewhere
* Notification at the top when user is offline
* ProfileSettings, when saving avatar
* In SecuritySettings
  * while generating 2fa QR code
  * while validating TOTP code
* In all UserPickers, when searching for a user

----
### Repository owner


suprko:
@anri-asaturov this is not urgent, just converts the circular ProgressBar back to SVG from pure CSS. It makes the loading spinner look a little nicer in a couple places but otherwise is mostly just to make the component easier to use. Up to you if you want to merge.

PR #244, opened 2018-02-12
 fix: room invite routing broke 'create dm/room' buttons in zerochats

suprko:
https://app.clubhouse.io/peerio/story/3296/desktop-new-room-and-new-dm-buttons-broken-on-zero-chat-screen
 
#### Testing instructions  
Should be fixed:
* ZeroChats "Create Room" and "Create DM" buttons

Hypothetically affected things, but should not be newly broken:
* Routing to ZeroChats when clicking on "chats" button in nav and user has no received chats or invites
* Room invites 

#### Notes
There was some weird routing going on... basically the new room invites were originally developed (by me) in a way that was basically entirely inside the ChatList component, because that was the only way to implement them on client only. Sam added some SDK support, but there were still some leftover routes in ChatList. The bug was caused by a route in ChatList that would send user to ZeroChats if user had no rooms/invites/DMs; however, ChatList is there when user creates a room or DM, so this meant that ZeroChats view could not be escaped.

----
### Repository owner


ebilg:
testing

ebilg:
looks good

PR #249, opened 2018-02-21
fix: wasn't able to activate other room invites after declining one

suprko:
https://app.clubhouse.io/peerio/story/3664/desktop-unable-to-view-invite-after-declining-a-different-one
 
#### Testing instructions  
Clicking other invites after declining an invite should work now.

----
### Repository owner


suprko:
@anri-asaturov @seavan found a way that works but not sure if it's the best way. Uses a `reaction` to watch the `chatInviteStore`. Another way I considered was for "reject invite" to be another view & route but that seems more complicated?

src/ui/chat/components/ChannelInvite.jsx
```diff
@@ -33,9 +33,18 @@ class ChannelInvite extends React.Component {
 
     @action.bound rejectInvite() {
         this.declined = true;
+        this.disposer = reaction(() => chatInviteStore.activeInvite, activeInvite => {
+            if (activeInvite && this.declined === true) {
+                this.declined = false;
+            }
+        });
```
anri-asaturov: ouch. so, what's going on in here? Same component renders different actual invites?

anri-asaturov:
Yeah this is quite hacky and unobvious, I'll think about it for a bit.

src/ui/chat/components/ChannelInvite.jsx
suprko: Yeah. `ChannelInvite` is its own view. We route to it when an invite is "activated".

When user rejects an invite, we set `declined=true`, which is what shows the "invite declined" screen. When user selects another invite, that invite is activated but `declined=true` persists.

The trouble is that selecting another invite happens in `ChatList`, which doesn't have access to `declined`. So this reaction listens for a change in the currently activated invite and sets `declined=false` if another invite is activated.

As I say in 1st comment a less hacky (but more complicated?) way is to have a whole new component like `InviteDeclined`, and route to that when invite is rejected. :weary:

seavan: @anri-asaturov easiest way to solve it is to add @observable rejected for invite and "delayReject()" function which sets rejected to true, but removes the item 1000 ms later, so that UX or animation can still happen

anri-asaturov: I'll do that now. Change this to use activeInvite.declined instead of this.declined

anri-asaturov: no need for delayReject, delay will be the default behaviour

anri-asaturov: Done https://github.com/PeerioTechnologies/peerio-icebear/commit/07a66eb500c899029d034380ba36829811222c09

sdk version will be released shortly

suprko: @anri-asaturov not sure what's going on... `activeInvite` is not given a `declined`. Icebear up to date and linked.
screenshot from 2018-02-21 14 18 00

anri-asaturov: fixing

anri-asaturov: fixed

suprko:
Works now using `chatINviteStire.activeInvite.declined`

ebilg:
can be merged

PR #264, opened 2018-03-26
feat: handle update failures

dchest:
#### Relevant info and issue

When starting, Peerio checks if the last update failed (this is designed
to happen when starting a new version right after the update), and if
so, it will show the error to the user and ask them to try the update
again, download the new version manually, or continue use the current
(older) version.

This also changes the "update is ready to install" dialog from alert()
to PeerUI Dialog and introduces another dialog to indicate that the
update is in progress.

Languages are now loaded before mounting the root component to avoid
race condition when something may have been showed before translations
are initialized.

#### PR links 

Copy: https://github.com/PeerioTechnologies/peerio-icebear/pull/169

#### Screenshots

When update fails and Peerio launches, the user is presented this dialog (also, autologin is not performed):

screen shot 2018-03-26 at 17 27 17

* When clicking on "Continue using older version", the dialog disappears and update info cleanup is performed, so that this dialog won't be shown again.
* When clicking on "Download manually", Peerio opens a web browser with the download page, performs cleanup, and quits.
* When clicking "Try again", Peerio will try to install the update again (this may involve re-downloading it) and shows the following status dialog permanently until it quits:

screen shot 2018-03-26 at 14 50 14

(in case of error, it will restart and will again show the failed update dialog).

---

Unrelated to update failures, there's also a new dialog asking to install update instead of `alert()` that was shown before.  This dialog is temporary, just retaining the current behavior, until a better update workflow is implemented from this user story https://app.clubhouse.io/peerio/story/2097/desktop-auto-download-and-install-update-on-app-reload:

screen shot 2018-03-26 at 18 33 19

Clicking on OK now shows the same update status as in the screenshot above so that users aren't kept guessing what's going on.
 
#### Testing instructions  

The whole update flow needs re-testing, so please 2 staging releases again.

XXX: need to figure out how to test failures.

----
### Repository owner


dchest:
Maybe delay testing and merging this until after we release all the shared files stuff and implement silent updates? https://app.clubhouse.io/peerio/story/2097/desktop-auto-download-and-install-update-on-app-reload This way we can test the complete update thing rather than re-testing it again.

dchest:
Closing in favor of https://github.com/PeerioTechnologies/peerio-desktop/pull/287

PR #272, opened 2018-04-13
feat: jump-to-bottom button

suprko:
 https://app.clubhouse.io/peerio/story/4850/desktop-jump-to-most-recent-message
https://github.com/PeerioTechnologies/peerio-icebear/pull/177

#### Testing instructions  
When user has a chat open + user has scrolled up any distance from the bottom + user receives new messages in that chat:
* jump-to-bottom button with unread message counter should appear just above message input
* jump-to-bottom button and unread count should persist until user has scrolled to bottom, either manually or by pressing the jump-to-bottom button

----
### Repository owner


seavan:
@suprko I think you can implement the remaining part in the SDK, too.

what you need to do:

- there's an observable in desktop, which prevents autoscrolling when the chat view is scrolled far enough away
https://github.com/PeerioTechnologies/peerio-desktop/blob/dev/src/ui/chat/components/MessageList.jsx#L128

- this observable should be moved to clientApp and reset on component MessagesList dismount

- when chat is changed, reset this flag:
https://github.com/PeerioTechnologies/peerio-icebear/blob/feat-app-label/src/models/chats/chat.js#L47

- when the flag is set, this reaction should not markAllAsSeen:

https://github.com/PeerioTechnologies/peerio-icebear/blob/feat-app-label/src/models/chats/chat.message-handler.js#L25

i think that should be enough to finish the task.

suprko:
@seavan I've got this working. In your last comment you referenced things in `feat-app-label` branch which I can't view now 😆 

Here's what has been done:
* desktop MessageList `stickToBottom` has been moved to icebear clientApp `messageStickToBottom`
  * desktop changes the flag the same way as before
  * `messageStickToBottom=true` whenever a new MessageList is mounted (change chat)
* when `!messageStickToBottom`, icebear chat.message-handler does not `markAllAsRead`. unread count (and jump-to-bottom button) are persistent until users scrolls to bottom, either manually or with button

Nothing has changed in chat.js because I wasn't sure what flag is supposed to be reset...

seavan:
@suprko feat-app-label was merged to dev.

checked code, looks good to me. passing on to testing.

src/ui/chat/ChatView.jsx
```diff
@@ -240,6 +240,28 @@ class ChatView extends React.Component {
             <ChatSideBar open={uiStore.prefs.chatSideBarIsOpen} />;
     }
 
+    @observable messageListRef;
```
anri-asaturov: But we already have refs and scrolling methods (including smooth scroll) inside MessageList, why did you duplicate it here?

src/ui/chat/components/MessageInput.jsx
```diff
@@ -112,6 +113,31 @@ class MessageInput extends React.Component {
...
+    }
+
+    @computed get snackbarVisible() {
+        if (this.snackbarRef) return this.snackbarRef.isVisible;
+        return false;
+    }
+
+    renderJumpToBottom() {
+        const chat = chatStore.activeChat;
+        if (chat.unreadCount === 0 || clientApp.messageStickToBottom) return null;
```
anri-asaturov: Why don't we let user quickly go back to the last messages even if there are no new messages? I scrolled to far back in history and want to go to most recent, sounds like a reasonable case?

anri-asaturov:
This is not ready to test yet

src/ui/chat/components/MessageInput.jsx
suprko: 🤷‍♂️ ask design/Skylar

anri-asaturov: Pauls says it should be as i said

src/ui/chat/ChatView.jsx
suprko: The button itself needs to be placed in MessageInput for positioning reasons, whereas the scrolling method is in MessageList. So the scrolling method is passed to `MessageInput` as a prop in ChatView.

The issue is described in greater detail in this comment. tl;dr, we can put the button in MessageList itself, but then the positioning would need to calculate MessageInput height (which can increase with long messages).
https://app.clubhouse.io/peerio/story/4850/desktop-jump-to-most-recent-message#activity-5635

anri-asaturov: okay

ebilg:
looks good, can be merged

PR #273, opened 2018-04-13
fix: upload dialog share with name

suprko:
 https://app.clubhouse.io/peerio/story/5684/desktop-in-multi-upload-dialog-in-dm-share-with-line-sometimes-shows-wrong-username-shows-self-instead-of-dm-recipient

#### Testing instructions  
In a DM use drag-and-drop or "Upload and share file". In the dialog that shows up, the "Share with" line should display the recipient's name and username, not yours. (On prod, in some cases, this was showing self's name.)

#### More details
If there was no `targetContact` identified, the code was using `targetChat.participantUsernames[0]`, which just assumes that the other user will be in the 0th position instead of self? Or maybe we changed icebear and a previous version of `participantUsernames` didn't include self? Anyway this should work now and be the slightest bit more performant too (no `contactStore` lookup).

----
### Repository owner


ebilg:
can be merged

anri-asaturov:
@ebilg i'm sure it fails if you create a channel without participants (just you)

ebilg:
uploading to dm with self fails

PR #283, opened 2018-05-02
fix: unread message behaviour on app/chat focus

suprko:
https://app.clubhouse.io/peerio/story/4850/desktop-jump-to-most-recent-message
https://github.com/PeerioTechnologies/peerio-icebear/pull/187

#### Testing instructions  
Messages should no longer be getting marked as read when user is not scrolled to bottom of chat.
ChatList unread badge should not appear on an active chat.

----
### Repository owner


src/ui/chat/components/MessageList.jsx
```diff
@@ -97,7 +97,6 @@ class MessageList extends React.Component {
 
     scrollToBottomIfNeeded() {
         if (!chatStore.activeChat) {
-            clientApp.isReadingNewestMessages = true;
```
seavan: I don't understand why remove it here, can you clarify please?

suprko: Hmm maybe I misunderstand the function. What's the purpose of setting `isReadingNewestMessages` to true if activeChat doesn't exist?

anri-asaturov: To reset it for the next time, when it does exist

suprko: It already gets set to true on componentWillMount though... What's the situation where MessageList will be mounted without an activeChat, and then a chat will be activated without MessageList being remounted?

Anyway I can add it back. Was trying to see where a stray `isReadingNewestMessages = true` might be causing Eren's bug, but he has reported that this didn't fix it anyway so 🤷‍♂️

PR #284, opened 2018-05-04
WIP: new folder share-in-progress UI

suprko:
https://app.clubhouse.io/peerio/story/3578/desktop-adequately-communicate-to-user-about-folder-sharing-in-progress
 
#### Details
This is the markup for the story above. See the Figma for mocks -- note that you have to go to the second page in Figma titled "don't lock" (I had no idea until yesterday that Figmas could have multiple pages).

What is intended now is that, when you share a folder, a second line appears below that FolderLine (on hover) with "files in queue to be shared...". You can click the folder name to go to the **volume**, or you can click the "files in queue" to go to the **non-shared folder**.

In the **volume**, there is a new subheader telling you that this volume is in the process of being created. Also, there are placeholder FileLines representing the incoming files that are in the process of being converted.

In the **non-shared folder**, there is a similar subheader telling you how many files are in the process of being converted to shared. FileLines disappear as they are converted to shared, and then there will be a zero-files state (not implemented yet).

New variables in Files.jsx:
* `incomingFiles`: array of files still to be converted
* `isVolumeInProgress`: are we viewing the newly-created volume
* `isConvertingToVolume`: are we viewing the non-shared folder that is being converted to volume

Right now everything is just mock values. `newfs` is in the middle of a refactor so sharing a folder just makes it disappear. But once it's back I will probably need help setting up the routing.

The biggest change is that the FileLine and FolderLine css has been massively refactored, moved from _all.scss to a new _FileFolderLineShared.scss file. This was necessary because the new "files in queue" line broke the existing `.row` elements in FolderLine.

----
### Repository owner


seavan:
leaving it to @anri-asaturov

PR #285, opened 2018-05-07
feat: DM placeholder when invite accepted

suprko:
clubhouse: https://app.clubhouse.io/peerio/story/5590/desktop-placeholder-dm-for-new-contacts
icebear:  https://github.com/PeerioTechnologies/peerio-icebear/pull/189

#### Testing instructions  
See CH story for the Figma mockups. I found the flow kind of annoying to follow on Figma so here's a summary:
* `existinguser` invites `newuser` via email
* `newuser` signs up and confirms email address
* now both `existinguser` and `newuser` should have a "placeholder" DM in their ChatList
  * both have same design but with slightly different text
* this placeholder DM has two buttons:
  * dismiss -> placeholder DM removed from ChatList
  * message -> placeholder DM removed from ChatList and replaced with real DM
    * instead of the usual chat header "This is the beginning of your chat history...", newly created DM shows the same text as the placeholder DM, until app restart
* if pending DM exists and DM with that user is created through another means (e.g. "Direct messages +" button), this should dismiss the pending DM

**There was also a refactor of the "plus icon" component**, which affects several other views. These should look unchanged:
* ChatList: "Rooms +" and "Direct messages +" buttons
* Contacts view: "Contacts +" button
* ZeroChats: "+ Create a room" and "+ Create a direct message" buttons
* Onboarding: "Create a room" button
  * this one was using the .chat-item-add class but then hiding the icon, I'm not sure why. So I just removed the class.
----
### Repository owner


src/ui/chat/components/PendingDM.jsx
```diff
@@ -0,0 +1,64 @@
...
+        chatStore.activeChat.dismiss();
+    }
+
+    onMessage = () => {
+        chatStore.activeChat.start();
+        if (this.props.onMessage) this.props.onMessage();
+    }
+
+    render() {
+        if (!chatStore.activeChat) return null;
```
anri-asaturov: also check for `chatStore.activeChat.isInvite`, we don't want to render this accidentally for normal chat instances.

src/ui/chat/components/PendingDM.jsx
```diff
@@ -0,0 +1,64 @@
...
+        const c = chatStore.activeChat.contact || chatStore.activeChat.otherParticipants[0];
```
anri-asaturov: according to sdk chatStore.activeChat.contact always exists if otherParticipants[0] does

anri-asaturov:
Not ready for test yet, also sdk part is not approved yet.

src/ui/shared-components/PlusIcon.jsx
```diff
@@ -0,0 +1,12 @@
+// TODO: use this component in ChatList and ZeroChats
+const React = require('react');
+
+class PlusIcon extends React.PureComponent {
+    render() {
+        return (
+            <span className="plus-icon">+</span>
```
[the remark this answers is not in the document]
suprko: This is already used several places and right now it's just a div with `chat-item-add` class, and then `chat-item-add.scss` is its own file in src/styles/components, which is... clunky. It's used often enough that making it a component makes sense IMO.

For consistency I can also replace all the other instances of `chat-item-add` divs if that's preferred

src/ui/chat/components/PendingDM.jsx
```diff
@@ -0,0 +1,64 @@
+const React = require('react');
+const { observer } = require('mobx-react');
+
```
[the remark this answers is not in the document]
suprko: Ehhhhhh this is a habit I got into after PeerUI. I just found it easier to read when the imports are separated into core libraries + icebear stores/helpers/etc + UI components. buuut I can erase them if they're ugly.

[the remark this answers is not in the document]
anri-asaturov: just wanted to know the principle

src/ui/chat/components/PendingDM.jsx
suprko: which file in SDK shows this?

anri-asaturov: https://github.com/PeerioTechnologies/peerio-icebear/pull/189/files#diff-4b1af765d9a7c3724131c298ec80082eR25

here, you see, it returns a contact from this.username, which is filled in constructor. If someone passed no value to constructor that's not normal, we don't account for that.

suprko:
@anri-asaturov updated
1. PendingDM checks for `activeChat.isInvite`
1. PendingDM no longer checks for `otherParticipants` when it already chekcs for `activeChat.contact`
1. Weird `.chat-item-add` divs replaced with PlusIcon component throughout the app
1. EmojiImage component uses emoji.js helper `emojiByCanonicalShortname`

suprko:
@anri-asaturov in PendingDM had to go back to the old `activeChat` checks at the top of `render`, to address Eren's bug. here's why:

https://github.com/PeerioTechnologies/peerio-desktop/blob/c00e823211a7f5e079d1bc7589c2a36a8d2ec028/src/ui/chat/components/MessageList.jsx#L213

PendingDM is not its own view but is a component used in ChatView and in MessageList. (If you click "message" button from PendingDM, it creates a DM using the PendingDM text in the place of the usual chat header). If we return null when `!isInvite`, this doesn't render. Also, `chatStore.activeChat.contact` doesn't exist for regular chat, just for PendingDM, so we still need `chatStore.activeChat.otherParticipants`.

src/ui/chat/components/MessageList.jsx
```diff
@@ -208,6 +209,18 @@ class MessageList extends React.Component {
 
     renderChatStart() {
         const chat = chatStore.activeChat;
+
+        if (chat.isChatCreatedFromPendingDM) {
+            return (
+                <div className="messages-start">
+                    <PendingDMHeader
+                        isNewUser={chat.isReceived || chat.isNewUserFromInvite}
```
seavan: because it's always Chat instance here, not ChatPendingDM, the property name is always isNewUserFromInvite

src/ui/chat/components/PendingDM.jsx
```diff
@@ -0,0 +1,48 @@
...
+    render() {
+        if (!chatStore.activeChat) return null;
+        const c = chatStore.activeChat.contact || chatStore.activeChat.otherParticipants[0];
+        if (!c) return null;
+
+        return (
+            <div className="pending-dm">
+                <PendingDMHeader
+                    isNewUser={chatStore.activeChat.isReceived || chatStore.activeChat.isNewUserFromInvite}
```
seavan: here it's always gonna be isReceived (because it is ChatPendingDM)

ebilg:
can be merged pending changes requested by Samvel

PR #291, opened 2018-05-18
feat: Medcryptor styles

alexandra-marin:
#### Relevant info and issue/PR links 
 https://app.clubhouse.io/peerio/story/4584/bootstrap-medcryptor-desktop-app
MC styles: https://github.com/PeerioTechnologies/medcryptor-desktop/commit/ec51f43c24de7088c58912e3a16e55dda43f5381
#### Testing instructions  

----
### Repository owner

- [ ] Was tested and can be merged.
- [ ] PR name follows conventional changelog format.

[PR guideline](https://github.com/PeerioTechnologies/peerio-icebear/blob/dev/docs/CONTRIBUTING.md)

PR #295, opened 2018-05-24
chore: extract peer-ui

suprko:
https://app.clubhouse.io/peerio/story/6624/extract-peerui-into-a-self-contained-repo

PeerUI repo itself: https://github.com/PeerioTechnologies/ui-library
Note to self to remember to incorporate these changes: https://github.com/PeerioTechnologies/peerio-desktop/pull/291

#### Testing instructions  
At some level this is similar to the time we replaced React Toolbox with PeerUI: technically the entire app is affected, but nothing should look or act any different. This time *should* be simpler, because we haven't changed any components at the functional level -- we've just translated them to TypeScript, which should actually make errors less likely since it's a stricter language. Here's hoping.

----
### Repository owner


src/styles/font.scss
```diff
@@ -0,0 +1 @@
+@import './static/fonts/OpenSans/opensans.css';
```
suprko: Dealing with font import was a nightmare, so PeerUI is now fontface agnostic and relies on consumer to manage font CSS and font files themselves.

src/styles/main.scss
```diff
@@ -5,19 +5,20 @@
 @import './static/emoji/sprites/emojione.css';
 
 // Brand/identity
-@import './vars/_branding';
+@import 'node_modules/peer-ui/src/scss/_branding';
```
suprko: obviously this kind of filepath is not ideal, what do you think?

suprko:
arg. still have to deal with Avatar onclick thing

src/styles/main.scss
anri-asaturov: this filepath is fine

suprko:
@anri-asaturov reopened, works for real this time. uiStore ContactProfile trigger thing replaced with new AvatarWithPopup component

package.json
```diff
@@ -111,6 +111,7 @@
     "node-sass": "4.7.2",
     "npm-run-all": "4.1.2",
     "opt-cli": "1.6.0",
+    "peer-ui": "github:PeerioTechnologies/ui-library#master",
```
anri-asaturov: Ideally we need releases, and semver, make a task for that

suprko: https://app.clubhouse.io/peerio/story/7807/peerui-releases-and-semver

lostfictions: @suprko did you review whether other deps become unused with this change?

lostfictions: Also looks like the peerui repo lists classnames as a devDep rather than a regular dependency... won't that break things (unless it's an implicit/transitive dependency in here....?)

suprko: Moved `classnames` to regular deps in PeerUI. As for desktop deps becoming unused, I've just checked now, doesn't look like there is anything.

PR #296, opened 2018-05-24
style: fix SignupProgress indicator and Files search bar

suprko:
https://app.clubhouse.io/peerio/story/1543/desktop-carousel-progress-point
https://app.clubhouse.io/peerio/story/4441/desktop-files-search-font-size

#### Testing instructions
* Files search bar font size is now default `input` font size
* SignupProgress indicator now starts on 1st dot (on "Welcome to Peerio" screen).
  * Account Key reminder prompt reuses 3rd dot

----
### Repository owner


src/ui/signup/SignupProgress.jsx
```diff
@@ -5,9 +5,13 @@ const { observer } = require('mobx-react');
 @observer class SignupProgress extends Component {
     render() {
         const { step, count } = this.props;
+        console.log(`${step} / ${count}`);
         const items = [];
         for (let i = 1; i <= count; ++i) {
-            const current = i === step ? <div className="current-step" /> : null;
+            const current = (i - 1 === step) || (i === step && step === count)
```
anri-asaturov: Could you explain a bit what do you mean by this? 
Especially (i === step && step === count) which, basically means `i === count`

suprko: There are four steps but only three "dot" indicators. On `dev` it goes from no dot highlighted, then 1st, then 2nd, then 3rd. Design requested that it starts on 1st dot highlighted, 2nd, 3rd, and then on the last step the 3rd one remains highlighted.

If we only put `i === count` then the 3rd dot will always be highlighted, because `current` will always return true for the 3rd dot even if the user has not navigated to the 3rd step.

ebilg:
can be merged

PR #299, opened 2018-05-25
feat: UserPicker add appLabel filter

suprko:
https://app.clubhouse.io/peerio/story/7810/desktop-make-user-pickers-filterable-by-applabel

#### Testing instructions  
n/a for now, this is just a chore to prepare for different user search behaviour on peerio vs medcryptor

----
### Repository owner


src/ui/shared-components/UserPicker.jsx
```diff
@@ -135,21 +135,40 @@ class UserPicker extends React.Component {
...
-            this.foundContact = !c.notFound && c;
             this.contactLoading = false;
+
+            /*
+                Restricting search by appLabel.
+                appLabel prop may be a single string (e.g. 'peerio')
+                or an array of strings (e.g. ['peerio', 'medcryptor'])
+            */
+            const { appLabel } = this.props;
+            if (appLabel && (
```
suprko: @flohdot is this alright? is this terrible?

I basically want to make it easy to pass either a single string as appLabel (e.g. 'peerio'), or an array (e.g. ['peerio', 'medcryptor']), so that we can handle all of the weird combinations that Medcryptor wants.

anri-asaturov: @seavan  suggested to extract filter to whitelabel logic file and call it here
so that all applabel conditions go there

suprko: @anri-asaturov yes that's for distinct whitelabel vs. Peerio components. In this case `appLabel` refers to a prop that is passed to `UserPicker` component that decides whether or not, to filter out users during user search. (This `appLabel` is not the same as `config.appLabel`)

The reason being that Medcryptor *sometimes* needs to search both MC and Peerio users, but sometimes not. https://app.clubhouse.io/peerio/story/7812/mc-desktop-filter-user-pickers-by-applabel

seavan: @suprko better call this property "context" and pass it on to the filtering function. filtering function will decide itself what to return.

don't call it appLabel though, it's confusing :)

suprko:
It's just a convenience thing. As long as people know that they have to pass an array even for just 1 appLabel then it's fine.

src/ui/shared-components/UserPicker.jsx
```diff
@@ -135,21 +135,40 @@ class UserPicker extends React.Component {
...
+
+            /*
+                Restricting search by app context.
+                `context` prop may be a single string (e.g. 'peerio')
+                or an array of strings (e.g. ['peerio', 'medcryptor'])
+            */
+            const { context } = this.props;
+            if (context && (
+                (typeof context === 'string' && c.appLabel !== context)
+                || (Array.isArray(context) && !context.includes(c.appLabel))
```
anri-asaturov: How and why can context be an array?

flohdot: let's go with always an array

PR #301, opened 2018-05-28
fix: reaction disposer in MigrationDialog

suprko:
n/a

#### Testing instructions  
should not get `TypeError: Cannot read property 'forEach' of undefined` error when MigrationDialog unmounts

----
### Repository owner

### DIFF
--- a/src/ui/shared-components/MigrationDialog.jsx
+++ b/src/ui/shared-components/MigrationDialog.jsx
@@ -116,9 +116,9 @@ class MigrationDialog extends React.Component {
 
     render() {
         // don't show migration progress if user has no shared files
-        // if (fileStore.migration.started && !fileStore.migration.hasLegacySharedFiles) {
-        //     return null;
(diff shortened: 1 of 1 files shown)


src/ui/shared-components/MigrationDialog.jsx
```diff
@@ -118,17 +116,17 @@ class MigrationDialog extends React.Component {
 
     render() {
         // don't show migration progress if user has no shared files
-        if (fileStore.migration.started && !fileStore.migration.hasLegacySharedFiles) {
-            return null;
-        }
+        // if (fileStore.migration.started && !fileStore.migration.hasLegacySharedFiles) {
+        //     return null;
+        // }
```
anri-asaturov: So we always render migration dialog now?

suprko: 🤦‍♂️ wups

PR #304, opened 2018-05-29
WIP: Patient space

suprko:
https://app.clubhouse.io/peerio/story/7820/desktop-patient-space-room-sidebar-markup
https://app.clubhouse.io/peerio/story/8141/mc-desktop-markup-for-patient-space-zero-state
https://app.clubhouse.io/peerio/story/7831/desktop-patient-room-styles-markup

#### Testing instructions  

----
### Repository owner


flohdot:
currently conflicts with `medcryptor-base`, fyi

suprko:
For sure can have more handling via activeSpace than route, but a route to the "patients" space does need to exist in some places because the PatientZeroChats screen contains important information (moreso than the regular ZeroChats) and it's pretty important for the user to see it.

src/ui/chat/components/ChatList.jsx
```diff
@@ -320,6 +357,26 @@ class ChatList extends React.Component {
                                     {this.allRoomsMap}
                                 </FlipMove>
                             </List>
+
+                            {config.whiteLabel.name === 'medcryptor'
+                                ? <List>
```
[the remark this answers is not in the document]
suprko: even for such a huge component like ChatList? Vast majority of it will be copy paste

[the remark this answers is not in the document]
anri-asaturov: not sure I understand, you have a condition `config.whiteLabel.name === 'medcryptor'` and everything below is a medcryptor ui, we agreed on extracting medcryptor ui into separate components, in this case, peerio counterpart will be an empty div.

src/whitelabel/components/PatientList.jsx
```diff
@@ -0,0 +1,14 @@
+const React = require('react');
+const config = require('~/config');
+const PatientListComponent = require('~/ui/whitelabel/medcryptor/PatientList');
+
+class PatientList extends React.Component {
+    render() {
```
suprko: @anri-asaturov @seavan @flohdot 
OK tell me what you think of this... I didn't want to make a PatientList component for Peerio that just returns an empty div, because I thought that was just unnecessarily cluttering up the folders. So I thought I'd just put the empty div in the `whitelabel/components/PatientList` handler itself.

But **now** I'm thinking... since we've moved all of the WLs into separate files anyway, why not put the entire WL file here? Instead of having UserSearchError_medcryptor in the same folder as UserSearchError we could just put all of the WL files in one place.

app/package.json
```diff
@@ -103,4 +103,4 @@
     "sanitize-filename": "1.6.1",
     "zxcvbn": "4.4.2"
   }
-}
+}
```
anri-asaturov: you've reconfigured your editor? this file is not validated by linter, so better bring newlines back

src/whitelabel/components/PatientList.jsx
anri-asaturov: > I didn't want to make a PatientList component for Peerio that just returns an empty div 
it's totally fine to do that, in fact it's better than any alternative

anri-asaturov: I think it's better to keep WL and Peerio variations together because when you edit one - you will notice there's another.

PR #305, opened 2018-05-29
fix: UserPicker use namespace context for user search

suprko:
https://app.clubhouse.io/peerio/story/7810/desktop-make-user-pickers-filterable-by-applabel

#### Testing instructions  

----
### Repository owner


suprko:
@seavan @flohdot could use some advice...

`getContactInContext` function (in `whitelabel/functions.js`) was cool but did not work in reality. Because:

* `const c = contactStore.getContact(query)` can return a contact object that is `loading`
* we use `c.loading` downstream (in UserPicker) for certain UI behaviours
* when `c.loading`, the other props for `c` (like `appLabel`) return `undefined`. so, we can't compare `c.appLabel` to `context` until `!c.loading`

I couldn't think of a way that would allow `getContactInContext` to return a `loading` contact object **and** allow it to compare `appLabel` against the namespace. So for now I just simplified the function to something that compares an existing contact object with `context`, and called the function in UserPicker inside the `when !c.loading` block.

It's less elegant but it works... let me know if you think there's a better way?

src/whitelabel/functions.js
```diff
@@ -1,16 +1,22 @@
...
-function getContactInContext(query, context) {
-    const c = contactStore.getContact(query);
-
-    if (context) {
-        if (context.includes(c.appLabel)) return c;
-        return { notFound: true };
+function checkContactInContext(contact, context) {
+    let c = contact;
+    if (context && !context.includes(contact.appLabel)) {
+        c = Object.assign({}, c);
```
suprko: In fact `Object.assign` is necessary. Here's why:

Let's say in MC we search for user in UserPicker in NewChannel. Peerio users are filtered out in NewChannel UserPicker, so let's say we set `user.notFound` is set to `true` as this function did before.

Then we go to NewChat, where both Peerio and MC users should be searchable. We search for that same user. Because `user.notFound=true`, this user has now disappeared from NewChat UserPicker, even though they should still be there in this view.

suprko: @seavan ^ part of our convo yesterday

seavan: problem with this code is that all observable properties are no longer observable after this cloning.

```
class Test {
    @observable notFound = false;
}
    testObservable() {
        const result = new Test();
        setTimeout(() => {
            result.notFound = true;
        }, 10000);
        when(() => result.notFound, () => {
            console.log('observable works');
        });
        const clone = Object.assign({}, result);
        setTimeout(() => {
            clone.notFound = true;
        }, 10000);
        when(() => clone.notFound, () => {
            console.log('clone observable works');
        });
    }
```

would not output success for a cloned one

suprko: How about not bothering with `notFound`, instead setting a new flag like `outOfAppContext` or something? It will get set again on each search so it shouldn't cause contact to disappear forever. And an added advantage is we have more control over the error message we return, e.g. we could say "username not available within this app. Please see this blog post for more info"

seavan: @suprko that would require to change everywhere dependency on notFound to that of new member. not nice.

I've moved your code to SDK, it looks something like this now:

```
    // Filter contacts in whitelabel namespace
    // Context corresponds to part of UI where filtering is applied
    // Supported contexts = default (null or undefined), and newchat
    // For Peerio, all contexts including newchat currently return only Peerio contacts
    // For Medcryptor, newchat context returns all namespace contacts, default context
    // returns only medcryptorw
    getContact(usernameOrEmail, context) {
        const result = new Contact(usernameOrEmail);
        const c = this.store.getContact(usernameOrEmail);
        // when our request is complete, check and apply filter
        when(() => !c.loading, action(() => {
            const filter = filters[config.whiteLabel.name || 'peerio'];
            if (filter(c, context)) {
                Object.assign(result, c);
            } else {
                result.notFound = true;
            }
            result.loading = false;
        }));
        return result;
    }
```

I'll wrap up desktop and push to medcryptor-base and tomorrow we can discuss and test.

suprko: I don't mean replace `notFound`, I mean we *also* have `outOfContext`. We'd need to update a couple of things in UserPicker but overall it wouldn't change much.

seavan: why add another property if it doesn't have any different ui representation so far? it'll just clutter the logic. especially if it is avoidable.

suprko: so you're not worried about the `Object.assign` new object no longer being observable?

seavan: @suprko look at my code. I instantiate a new Contact() instead of empty object. That way, all observables are in place (same as for the object we get through getContact).

anri-asaturov: Do you really need to modify getContact, please find a way around it, what's wrong with separate filtering, even if you add it to sdk

seavan:
ok, the code is moved to here 

https://github.com/PeerioTechnologies/peerio-icebear/commit/67055d88f0512355928fa4322e55417cb2860461

desktop usage is here:

https://github.com/PeerioTechnologies/peerio-desktop/commit/d231e8cb2e4266a3dcc3216b6e35f4bf0c8b8a29

closing this pull request.

src/whitelabel/functions.js
seavan: @anri-asaturov it's in module. in whitelabel.getContact. original getContact is not modified at all https://github.com/PeerioTechnologies/peerio-icebear/commit/67055d88f0512355928fa4322e55417cb2860461

PR #307, opened 2018-05-30
fix: allow creating folders in folder selection dialog

dchest:
#### Testing instructions  

----
### Repository owner


anri-asaturov:
Why would you create folder when picking files to upload?

dchest:
This is one is used in `downloadFolder`, not for uploading (that one is called `pickLocalFiles`).

dchest:
Also, the button label is `Open`, should we set it to `Save` or `Download`? In which case it's probably better to rename `selectFolder` to `selectFolderToDownloadInto` or something.

dchest:
^ made the changes, pls re-review

PR #325, opened 2018-06-08
fix: check chat.metaLoaded, fix empty DM titles

suprko:
https://app.clubhouse.io/peerio/story/8955/desktop-chatview-doesn-t-check-if-activechat-chathead-is-null-and-renders-and-crashes

#### Testing instructions  
* DM titles were empty. Receipient's name should be there again.
* Skylar got a blue screen of death due to a null `chatHead`. Wasn't able to repro but there is now a check so chats will not load if `metaLoaded` is false (so, can only load if `chatHead` exists)

----
### Repository owner


suprko:
@anri-asaturov I believe the issue is that I was using `ELEMENTS.chatEditor` (which looks for `chatHead`) where I wasn't supposed to. So I think this fixes it. However, should I add a null check anyway?

anri-asaturov:
You shouldn't render chats with metaLoaded === false. If you do that - you don't need any checks. I'm pretty sure it was the case before, we didn't render those chats. Where and how it got different - i don't know.

PR #345, opened 2018-06-17
style(SharedFolderActions): only show divider if there are items above & below it

suprko:
https://app.clubhouse.io/peerio/story/2762/desktop-unshare-a-folder-in-a-dm-from-more-vert-icon
 
#### Testing instructions  
This PR itself only fixes the styling of the "Unshare" button (from the menu on an inline folder that has been shared in a chat), but the entire functionality of this button still needs to be properly tested.

Since the button already exists on `dev` (just with an ugly line above it) this can probably be moved to Testing On Dev immediately...?

----
### Repository owner


ebilg:
can be merged

PR #349, opened 2018-06-20
feat: jump to bottom on message send

suprko:
 https://app.clubhouse.io/peerio/story/6873/desktop-autoscroll-when-sending-a-new-message-to-the-chat-yourself

#### Testing instructions  
Sending a message should cause the message list to scroll to the bottom.

----
### Repository owner


suprko:
Pretty simple but I don't know if this is the best way to implement this feature...?

anri-asaturov:
here

anri-asaturov:
Your way was not the best bcs 
1. it only worked when sending messages but not ACK or files
2. it used 'jump to bottom' that works a bit differently, sometimes it can reset the chat, which send will do anyway, and then there will be 2 resets

src/ui/chat/ChatView.jsx
```diff
@@ -71,32 +77,35 @@ class ChatView extends React.Component {
      * @param {Object} richText A ProseMirror document tree, in JSON.
      * @param {string} legacyText The rendered HTML of the rich text, for back-compat with older clients
      */
-    sendRichTextMessage(richText, legacyText) {
+    @action.bound sendRichTextMessage(richText, legacyText) {
         try {
+            this.scrollToBottom();
```
suprko: should use `jumpToBottom` here no? if we're scrolled up high enough that the newest chats have unloaded, we need `chat.reset` in order to actually go to the newest chat

anri-asaturov: send() in sdk does the reset if canGoDown

PR #352, opened 2018-06-26
fix: preserve line breaks in plain-text messages

dchest:
 
https://app.clubhouse.io/peerio/story/9382/plaintext-messages-don-t-have-line-breaks

#### Testing instructions  

Send the following message:

```
Hello world
http://clipart-library.com/img1/909785.jpg
```

(line breaks may be mangled by copying, so type out "Hello world", press Shift+Enter and then paste the link)

Check that the image preview appears and there's no `ERROR Error: worldhttp://clipart-library.com/img1/909785.jpg request failed` in console.

----
### Repository owner


lostfictions:
Looks like a good fix to me, thanks!

Do you think we should still consider switching to `innerText` eventually? I don't imagine more issues like this will keep cropping up.

dchest:
@lostfictions yeah, I tried various approaches, including `innerText`, but got tired of fighting with issues (like switching emoji to spans if we use `innerText` or automatically inserting new lines, which required changing CSS for message rendering), but decided this was the easiest approach for plaintext messages that we might remove in the future.

ebilg:
can be merged

PR #353, opened 2018-06-26
chore: PeerUI updates

suprko:
 https://app.clubhouse.io/peerio/story/5618/peerui-desktop-refactor-button-jsx-so-that-icons-inherit-colours-from-parent

#### Testing instructions  
IMO this can be merged straight to `dev`. I was testing the changes as I went along, and anyway the changes are kinda too vague and dispersed everywhere to be really testable in a specific way.

Bunch of chores/refactors:
* in PeerUI, Button icons inherit colour from parents, so on desktop I've removed a bunch of styles that were specifically defining `.material-icon` colour
* in PeerUI, I changed the Avatar `clickable` and `onClick` logic so that `clickable` no longer requires an `onClick`. also `onClick` now automatically applies the `clickable` class, so the only time we need `clickable` is when we want to apply the `clickable` styles without an `onClick` event
  * this happens a couple of places e.g. in MembersSection where clicking on the ListItem (rather than the Avatar itself) opens the ContactProfile popup, so Avatar doesn't have an `onClick` but we still want it to bounce on hover
* while doing ^these things I found a bunch of unused styles & classes and general disorganization, which have been fixed

----
### Repository owner


anri-asaturov:
ready to test

ebilg:
can be merged

PR #360, opened 2018-07-04
WIP: caching

anri-asaturov:


src/stores/indexed-db-storage.js
```diff
@@ -0,0 +1,117 @@
+class IndexedDbStorage {
+    constructor(name, keyPath) {
```
seavan: what do we use keyPath for?

src/stores/indexed-db-storage.js
```diff
@@ -0,0 +1,117 @@
...
+            request.onsuccess = function() {
+                resolve(request.result);
+            };
+        });
+    }
+
+    setValue(key, value, confirmUpdate) {
+        return new Promise((resolve, reject) => {
+            if (!confirmUpdate) {
+                const request = this.getStore().put(value);
```
seavan: why aren't you providing the key explicitly?
are you sure "value" is always going to have "key" to infer from?

src/stores/indexed-db-storage.js
anri-asaturov: it's just the way IndexedDB works, it wants to get keys out of objects.

src/stores/indexed-db-storage.js
anri-asaturov: yeah

PR #361, opened 2018-07-04
feat: redesigned UI for message send error actions

suprko:
https://app.clubhouse.io/peerio/story/9689/desktop-message-errors

#### Testing instructions  
Failed message (e.g. disconnect from internet and try to send a message) has been redesigned. Same functions, but they're more obvious now, instead of being in a hidden dropdown menu.

Note that failed inline files (via upload or share) are not included in this PR, contrary to the mocks. This is much bigger task and likely needs a different design. Paul will make a new story.

----
### Repository owner


src/ui/chat/components/Message.jsx
```diff
@@ -51,8 +52,8 @@ let legacyProcessMessageForDisplay;
 @observer
 class Message extends React.Component {
     @observable.shallow errorData = null;
-    @observable errorMenuVisible = false;
     @observable allowShowSendingState = false;
+    @observable resendInProgress = false;
```
anri-asaturov: why another flag and not m.sending ?

anri-asaturov: do you think we can reduce flag count?
what do we need to know? It seems too much

suprko: We want the red background (`.send-error`) to apply when you've had `m.sendError`. But when you click "Try again", the `send` function sets `m.sendError` to `false`. So we need another flag to continue showing the "errored" background color even when the message itself strictly speaking isn't errored anymore. Setting a local flag `resendInProgress` allows this.

`allowShowSendingState` was there already. This is because we don't want the "Sending..." message to appear every time a message send attempt is in progress -- this would cause a flash every single time you send a message. "Sending..." UI only appears if attempt has taken > 3 sec or if the message has previously `sendError`'d.

🤷‍♂️ it's a bit of a finicky design, idk if we can keep the requested functionality with fewer flags

ebilg:
can be merged

PR #364, opened 2018-07-04
feat: drag and drop between folders

lostfictions:
 
https://app.clubhouse.io/peerio/story/7210/desktop-drag-drop-between-folders
https://app.clubhouse.io/peerio/story/8507/drag-and-drop-flows
https://app.clubhouse.io/peerio/story/7394/document-drag-and-drop

Some extra info about design choices made: https://app.clubhouse.io/peerio/story/9510/shared-folders-plus-drag-drop-is-confusing

Copy PR: https://github.com/PeerioTechnologies/peerio-icebear/pull/248

#### Testing instructions  

- Drag a single item into a non-shared folder
- Drag multiple items into a non-shared folder
- Repeat both of the above with shared folders (should display a confirmation if the user hasn't selected to never warn about it)

~~~KNOWN ISSUES: dragging a file from the desktop into the Files pane might break after doing it once -- seems to be a conflict between the global drag and drop context and the files-specific one. To be verified~~~
Should be fixed.

----
### Repository owner


lostfictions:
Note that dragging and dropping into a shared folder doesn't seem to work now (but neither does moving into a shared folder via the bulk action buttons, so it's not specific to drag and drop).

ebilg:
can be merged

PR #365, opened 2018-07-05
fix: small Files UI updates

suprko:
 https://app.clubhouse.io/peerio/story/10731/desktop-check-that-recent-fixes-and-changes-on-files-got-included-in-big-files-refactor

#### Testing instructions  
Catching the small bugs that fell through the cracks in the big Files refactor.
* reenable the three-dots menu on fileline/folderline that has been selected
* on an individual file's three-dots menu, when you click share or move, the file should be shown as checked/selected under the dialog that pops up (and selection should clear on dialog close)

OK so, if you bulk select multiple files and then use the three-dots menu on one file to move/share, only that one file will be shown as selected (the others will be deselected). If you cancel the dialog instead of moving/sharing, this file will stay selected. Design is deciding whether we want to restore the previous selection state, but for now I think this can go ahead and they can create a new bug if they want to change it.

----
### Repository owner


src/ui/files/components/FileActions.jsx
```diff
@@ -58,13 +58,18 @@ class FileActions extends React.Component {
...
+        fileStore.clearSelection();
+        this.props.file.selected = true;
+
         const folder = this.props.file.folder || fileStore.folderStore.root;
         this.moveFileDialogRef.current.show(folder, this.props.file);
     }
 
     shareFile = async () => {
-        const contacts = await this.shareWithMultipleDialogRef.current.show(null, 'sharefiles');
+        fileStore.clearSelection();
```
lostfictions: Don't these changes also need to happen in FolderActions too, then? Folder lines can also be part of a multi-selection, after all.

Though... there'll be an issue there 'cause FolderActions are also in the breadcrumb header.

src/ui/shared-components/ShareWithMultipleDialog.jsx
```diff
@@ -116,11 +119,13 @@ class ShareWithMultipleDialog extends React.Component {
         this.shareContext = '';
         this.folderWithExistingShare = null;
         this.query = '';
+        fileStore.clearSelection();
         this.selectedUsers.clear();
     }
 
     @action.bound share() {
         this.resolver(this.selectedUsers.values());
+        fileStore.clearSelection();
```
lostfictions: this dialog probably shouldn't be introducing side effects like this unless we have a really good reason?  it's just built to return a list of contacts, so the caller should await it and perform the side effect itself afterwards if it really needs to (but i think all the share operations clear the selection afterwards anyway...?)

for the most part it's actually not files-specific -- iirc the only reason it even has a conception of a folder being passed in is so that it can examine the folder to see if it's shared so it can display the contacts it's already shared with.

suprko: Yeah I've changed it to be done at the menu level but that's introducing its own problems 😩

anri-asaturov:
fix tests

suprko:
fixed

ebilg:
can be merged

PR #366, opened 2018-07-06
fix: share-to-chat status bar

suprko:
https://app.clubhouse.io/peerio/story/10779/desktop-inline-file-upload-status-bar-fixes
 
#### Testing instructions  
Upload-to-chat progress bar:
* It's no longer possible for percentage to exceed 100 (fixed on icebear `dev`)
* Cancel button works now

----
### Repository owner


src/ui/chat/components/ShareToChatProgress.jsx
```diff
@@ -53,15 +53,19 @@ class ShareToChatProgress extends React.Component {
                                 ? `${progressPercentage}%`
                                 : queued
                                     ? null
-                                    : progress ? `${Math.floor(progress / progressMax * 100)}%` : '0%'
-                                // TODO: for some reason it's possible for progress > progressMax ??
+                                    : progress
+                                        // TODO: for some reason it's possible for progress > progressMax ??
```
suprko: Infinite ternary operator nesting! 🚀 🚀 🚀 

It's possible for `progress` to exceed `progressMax` and I couldn't figure out why. I tried looking at how the upload progress is calculated in icebear and couldn't really dig into what was going wrong... I suspect it has something to do with the chunking? Maybe miscounting the number of chunks? Not sure.

So, this is a hacky workaround that just shows "100%" when `progress > progressMax`. This is good enough for now because when I was testing, it seemed that the disparity was greatest for small files. For example, a 6B file was showing millions of percent at completion, whereas an 6MB file was showing 115%. So it seems unlikely that anyone will get stuck at 100% for very long because it becomes less likely the larger the file is.

anri-asaturov: No, it's not good enough. I'll debug it.

src/ui/chat/components/ShareToChatProgress.jsx
```diff
@@ -26,6 +19,13 @@ class ShareToChatProgress extends React.Component {
         const { progress, progressMax, progressPercentage } = item;
         const fileType = fileHelpers.getFileIconType(fileHelpers.getFileExtension(item.name));
 
+        // TODO: handle folder share cancel once available
+        const handleCancel = () => {
+            if (!item.isFolder && confirm(t('title_confirmCancelUpload'))) {
+                item.cancelUpload();
+            }
+        };
```
anri-asaturov: since when we started to define event handlers in render like this? :) 
This component renders with every progress update, potentially very often.

anri-asaturov: An an alternative you can create a function to create and cache handler for current item. Or create a small component.

anri-asaturov:
I've fixed it in SDK https://github.com/PeerioTechnologies/peerio-icebear/commit/0d48495b2166cd3adaeba1ce731efc1d511688e6

Progress was bigger than  progressMax under some circumstances.
Please, remove the hack that checks this issue from here, if it's broken in future we want to know it, rather then component hiding it from us.

Also flatten out the ternary nesting, it's easy to do, just add a few `if`'s before returning jsx

suprko:
fixed

ebilg:
can be merged

PR #373, opened 2018-07-12
feat: user awareness of text formatting

lostfictions:
 
https://app.clubhouse.io/peerio/story/122/desktop-user-awareness-of-text-formatting

#### Testing instructions  

----
### Repository owner


src/ui/chat/components/FormattingButton.jsx
```diff
@@ -0,0 +1,136 @@
...
+const {
+    EditorView // eslint-disable-line no-unused-vars, (used for typechecking)
+} = require('prosemirror-view');
+const {
+    MarkType // eslint-disable-line no-unused-vars, (used for typechecking)
+} = require('prosemirror-model');
+const {
+    Plugin,
+    Transaction, // eslint-disable-line no-unused-vars, (used for typechecking)
+    EditorState // eslint-disable-line no-unused-vars, (used for typechecking)
```
anri-asaturov: There's no way to make it look into .js files automatically?

anri-asaturov:
redy to test

src/ui/chat/components/FormattingButton.jsx
lostfictions: Real TypeScript will elide imports that are only used for typechecking, so in the transpile step they disappear anyway -- it works well for avoiding eg. namespace collisions, and there's no overhead in the runtime. The TypeScript compiler and the TypeScript eslint plugins are smart enough to know when imports are being used in this way, too. ts-check mode unfortunately isn't that smart. It's better than nothing, but it's got all these weird limitations and hacks. (You'll see elsewhere in this PR that it got confused by at least one other issue, but that sort of thing never happens to me with TypeScript proper.)

lostfictions: (Also, in actual TypeScript you actually need less imports and type annotations, ironically enough -- there aren't all these weird failure modes where types fail to flow, and you can write interfaces and aliases that are actually shareable instead of having all these duplicated, verbose, noisy annotations and typedefs everywhere.)

ebilg:
can be merged

PR #376, opened 2018-07-16
fix: FolderActions in Breadcrumb

suprko:
https://app.clubhouse.io/peerio/story/11174/desktop-breadcrumb-menu-options-not-working
broken by this PR https://github.com/PeerioTechnologies/peerio-desktop/commit/a67f65a8d5b2c2922c7c7441cae8a094fd723c33
 
#### Testing instructions  
"Move", "Share", "Rename" should all work in Breadcrumb FolderActions again

#### Notes
The issue is that the dialogs (`AddOrRenameDialog`, `MoveFileDialog`, `ShareWithMultipleDialog`) are included in the `FolderActions` component. When Move/Share/Rename is clicked, the folder is selected; when any items are selected, the `Breadcrumb` component hides the `FolderActions` component, therefore the dialogs don't show up.

Fixed this with a `noSelection` flag on `FolderActions`, which seems a little bit clunky, idk...? Maybe just needs a better name.

----
### Repository owner


lostfictions:
This is.... kind of sorta exactly why I was suggesting moving the side effects out of FolderActions :P

This new version might fix the bug at hand, but I think it's potentially worse than before -- it still has the side effect but now it's even harder to reason about. I know all these globs of global mutable state make it hard to avoid, but do you think there's a way that this can be done more cleanly?

lostfictions:
(If nothing else I might suggest either a docstring for the prop or a more explicit propname, since now I'm not actually sure why/how some embedding contexts need the action to set the selection and some don't... which I guess is one reason this bug slipped in in the first place.)

suprko:
I'm gonna see about mocking the "selection" state in FileLine/FolderLine, since it's purely cosmetic anyway

suprko:
OK I think this a bit less weird now.

lostfictions:
(I don't think there's actually a way to have editor support for docstrings on interface definitions in ts-in-js -- one of the million flaws with it -- but it's still helpful for people just reading the class. And if we ever transition to Real TS or whatever it'll already be in place and just work.)

ebilg:
can be merged

PR #377, opened 2018-07-17
feat: Login and Sign Up telemetry

suprko:
https://app.clubhouse.io/peerio/story/10470/desktop-mixpanel-sign-up-trackers
icebear https://github.com/PeerioTechnologies/peerio-icebear/pull/258
 
#### Testing instructions  
Telemetry on Login and Sign Up views. Testing being done by Kaoru.

----
### Repository owner


suprko:
I prioritized keeping the component code as clean as possible, and storing the telemetry logic/functions in their own files. There are a couple of places where component directly calls `telemetry.send()` but only for the very simplest of events. The rest of the time, component will call a function that's named in a telemetry helper file corresponding to that component (in this case, `login.js` and `signup.js`). Essentially I've added a telemetry functions middleman in order to keep telemetry functionality out of the components as much as possible. But the tradeoff is that maybe there's too much redundancy?

src/telemetry/components/login.js
```diff
@@ -0,0 +1,32 @@
...
+    onLoginWithEmail: () => {
+        return textInput(S.USERNAME, null, S.SIGN_IN, S.ERROR, 'Using @');
+    },
+
+    toggleAkVisibility: (visible) => {
+        return [
+            S.TOGGLE_VISIBILITY,
+            {
+                Item: S.ACCOUNT_KEY,
+                Visible: visible
```
anri-asaturov: why do props start with capital letters?

src/telemetry/index.js
```diff
@@ -0,0 +1,83 @@
...
+            Object.keys(obj).forEach(k => {
+                // Each item in this object is either the event we want to send, or a function that returns the event
+                // that we want to send based on an argument (e.g. if a property can change based on app state).
+                if (typeof obj[k] === 'function') {
+                    this[item][k] = (...arg) => {
+                        if (obj[k](...arg)) this.send(obj[k](...arg));
+                    };
+                } else {
+                    this[item][k] = () => {
+                        if (obj[k]) this.send(obj[k]);
```
anri-asaturov: `if (obj[k])` ? How can this happen? Why don't we skip creating the function?

src/telemetry/index.js
```diff
@@ -0,0 +1,83 @@
...
+        Object.keys(sharedObj).forEach(k => {
+            obj[k] = (...arg) => {
+                // If the function returns null, do not send the event
+                if (sharedObj[k](context, ...arg)) {
+                    this.send(sharedObj[k](context, ...arg));
+                }
+            };
+        });
+        return obj;
+    }
```
anri-asaturov: I didn't understand this part tbh, too many function and property generations, it's impossible to follow your thought.

anri-asaturov: Try maybe better naming, like instead of `obj` name it something what it really is?

src/telemetry/index.js
```diff
@@ -0,0 +1,83 @@
...
+        tm.init();
+    }
+
+    send(event) {
+        // Basic properties to send with all events.
+        // There are additional baseProps on SDK. These are desktop-specific (e.g. `window` doesn't exist on mobile)
+        const baseProps = {
+            'Window Width': window.innerWidth,
+            'Window Height': window.innerHeight
+        };
```
anri-asaturov: we're gonna send window dimensions with every event?

src/ui/login/Login.jsx
```diff
@@ -86,15 +87,22 @@ class LoginStore extends OrderedFormStore {
                         .catch(() => { this.loginStore.busy = false; });
                 } else this.loginStore.busy = false;
             });
+        this.startTime = Date.now();
+    }
+
+    componentWillUnmount() {
+        telemetry.login.duration(this.startTime);
```
anri-asaturov: we don't care if it was finished or not? it will get called even if user navigates back

src/ui/login/Login.jsx
```diff
@@ -86,15 +87,22 @@ class LoginStore extends OrderedFormStore {
...
 
 
     togglePasswordVisibility = () => {
         this.loginStore.passwordVisible = !this.loginStore.passwordVisible;
         if (this.akRef) this.akRef.inputRef.focus();
+        telemetry.login.toggleAkVisibility(this.loginStore.passwordVisible);
     };
 
     unsetLastUser = () => {
+        telemetry.send('Change User');
```
anri-asaturov: shouldn't this be a custom function call instead of direct 'send'?

src/ui/login/Login.jsx
```diff
@@ -148,6 +158,13 @@ class LoginStore extends OrderedFormStore {
         }
     };
 
+    usernameHandleKeyPress = (e) => {
+        if (e.key === '@') {
+            telemetry.login.onLoginWithEmail();
+        }
+        this.handleKeyPress(e);
+    }
```
anri-asaturov: what if they paste, or type @ accidentally and then delete? 
I think it's better to check the username when the form is submitted.

src/telemetry/components/login.js
suprko: Requested by Kaoru/marketing. They want all of the stuff in Mixpanel to be in Title Case

src/telemetry/index.js
suprko: Yes

src/ui/login/Login.jsx
suprko: Yep, also requested by Kaoru & Skylar. We send the time on page whether the action was completed or not. From other button clicks they can find out whether the action was completed.

src/telemetry/components/login.js
anri-asaturov: If there's no way to make a mapping, I would rather add key transform right before sending it to mixpanel instead of breaking people's eyes and code style like this.

src/ui/login/Login.jsx
suprko: Trust me I have argued a lot about this! Listening for the '@' key press is their specific request.

src/telemetry/components/login.js
anri-asaturov: We can even transform keys to something better like 'someProperty' => 'Some Property'

src/telemetry/index.js
anri-asaturov: Why? Listen to resize event and send new dimensions when needed.

suprko: I know, listen I have spent like a week arguing about this stuff 😩 if you have concerns about the events themselves, talk to Kaoru and Skylar.

src/telemetry/helpers.js
```diff
@@ -0,0 +1,100 @@
...
+    SAVED: 'Saved',
+    SIGN_IN: 'Sign In',
+    SIGN_UP: 'Sign Up',
+    START_ACCOUNT_CREATION: 'Start Account Creation',
+    TERMS_OF_SERVICE: 'Terms Of Service',
+    TEXT_INPUT: 'Text Input',
+    TOGGLE_VISIBILITY: 'Toggle Visibility',
+    USERNAME: 'Username',
+    VIEW_LINK: 'View Link'
+};
```
lostfictions: These look like several different contexts mixed together (locations, actions, items, etc.) -- would it make more sense to split each context out to make refactoring/amending/finding things more obvious?

src/telemetry/index.js
```diff
@@ -0,0 +1,89 @@
...
+                    };
+                } else {
+                    // If it's not a function then it's an array, and we can send the array directly
+                    this[item][k] = () => {
+                        this.send(obj[k]);
+                    };
+                }
+            });
+        });
+    }
```
lostfictions: I think I might've asked this already previously, but is it preferable to have telemetry be a class that wraps everything and passes them through this setup function, rather than externalizing the setup function and having each unit of telemetry bring in the setup function and export the result? The latter is slightly more verbose but might be less "clever", more flexible, and easier to reason about (and might typecheck better, and thus be more easily maintainable/refactorable.)

If there's good reasons to have a class (you really need a bag of global state, etc.) then maybe this approach is best, but I'm not totally sure I see the justification without knowing more.

src/telemetry/index.js
```diff
@@ -0,0 +1,89 @@
+const { computed } = require('mobx');
```
lostfictions: If we're moving to TypeScript soon, can you add `// @ts-check` annotations on new files and ensure they're well-typed?

src/telemetry/index.js
```diff
@@ -0,0 +1,89 @@
...
+        // Same as the event factory in constructor(), but includes `context` (corresponding to currentRoute)
+        // so the same component in multiple locations can send event with the correct location props
+        Object.keys(sharedObj).forEach(k => {
+            sharedEvents[k] = (...arg) => {
+                // If the function returns null, do not send the event
+                if (sharedObj[k](context, ...arg)) {
+                    this.send(sharedObj[k](context, ...arg));
+                }
+            };
+        });
```
lostfictions: These can probably just be memoized once instead of being in a computed getter? In general, the latter is brittle to accidentally introducing more frequently-changing observables and therefore causing recomputations to happen much more often than intended. Probably better to use a tiny bit more memory than to do work over and over.

src/telemetry/index.js
```diff
@@ -0,0 +1,89 @@
...
+        tm.init();
+    }
+
+    send(event) {
+        // Basic properties to send with all events.
+        // There are additional baseProps on SDK. These are desktop-specific (e.g. `window` doesn't exist on mobile)
+        const baseProps = {
+            windowWidth: window.innerWidth,
+            windowHeight: window.innerHeight
+        };
```
lostfictions: Have we talked about privacy concerns for fingerprinting by including this on every action?

src/telemetry/index.js
lostfictions: (I don't think this is necessarily a change that should be made -- mostly wondering if you tried/considered it already and thought these ergonomics were better, etc.)

lostfictions:
Looks a lot better than the previous iteration, btw! Thanks for all your work with this.

src/telemetry/index.js
suprko: Discussed with Skylar and Flo, they say it's OK

src/telemetry/index.js
suprko: This is done now and I do like it more; the "function factory" stuff is done in one less step of abstraction so it's easier to follow for sure.

I'm still not sure how to manage the `context` stuff in the shared components so `shared.js` is still exported as a singleton... I'll maybe ask you about it in a DM?

src/telemetry/helpers.js
lostfictions: I think it's okay to duplicate them and have a clear idea of usage sites, rather than make things confusing for the sake of having as little text as possible? Again, we can afford a few bytes of memorym and if we want to refactor later, split things out, etc., this might make it easier... I might not understand the issue though -- if you still feel this way is better that's fine too.

anri-asaturov:
plz fix conflicts

src/telemetry/components/signup.js
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        return [
+            S.NAVIGATE,
+            {
+                option: S.NEXT,
+                sublocation: sublocation || S.SIGN_UP
+            }
+        ];
+    },
```
seavan: this is actually kinda hard to read, @suprko.

you have a switch, which returns for 1 case, doesn't return for second case, but sets some value which is to be used later, but if it is not set, it defaults to some other value.

very error-prone to changes too.

why not move the step configuration in some constant and use it? something like this:

```
const ADVANCE_STEPS = [
    // we pressed create account
    [S.START_ACCOUNT_CREATION, { text: S.GET_STARTED }],
    [S.NAVIGATE, { option: S.NEXT, sublocation: S.SIGN_UP }],
    [S.NAVIGATE, { option: S.NEXT, sublocation: S.ACCOUNT_KEY }],
];

// ...

advanceButton: step => ADVANCE_STEPS[step];

```

same for retreat.

ebilg:
I think "WIP" can be removed from the PR name now

src/telemetry/main.js
```diff
@@ -0,0 +1,51 @@
...
+
+    return ret;
+}
+
+function makeSendFunction(item) {
+    // Each item in this object is either the event we want to send, or a function that returns the event
+    // that we want to send based on an argument (e.g. if a property can change based on app state).
+    if (typeof item === 'function') {
+        return (...arg) => {
+            if (item(...arg)) send(item(...arg));
```
anri-asaturov: cache result of item(...arg) call

src/telemetry/components/signup.js
```diff
@@ -0,0 +1,141 @@
...
+        return textInput(S.USERNAME, null, S.SIGN_UP, S.ERROR, errorMsg);
+    },
+
+    // AccountKey.jsx
+    copyAk: [
+        S.COPY,
+        {
+            item: S.ACCOUNT_KEY,
+            sublocation: S.ACCOUNT_KEY
+        }
```
anri-asaturov: can we further categorize strings in `S`? 
It's like if I would want to enumerate people's ages and weights I would create 'Numbers' enumeration.

src/telemetry/components/signup.js
suprko: Hummmm let me see what I can do. The tricky thing is that these are not the functions themselves, right? These are the objects that get passed to `main.js` where `setup()` turns them into `send()` functions. And `setup` doesn't currently have logic for creating an array-of-functions instead of individual functions. So idk, I'll figure out a way to make this more legible.

src/telemetry/components/signup.js
suprko: 🤔 Six asked about this too. My concern was that we would be doubling a lot of strings, e.g. "Sign In" can be a `location`, `action`, and `text`, so we'd be duplicating a lot. But if people want them categorized then yeah we can do that.

anri-asaturov: Any chance we MIGHT want to rename them separately eventually?

suprko: I suppose it's possible. Idk what Kaoru/Paul/Skylar's plans are.

src/telemetry/components/signup.js
seavan: @suprko this actually doesn't change anything, see:

```
const signup = setup({ 
    advanceButton: step => ADVANCE_STEPS[step],
// ...
});
```

seavan: you can skip the braces and return statement altogether @suprko , arrow function always returns its right-side value.

unless you plan to expand the body somehow.

ebilg:
Has this passed review? @lostfictions @anri-asaturov @seavan

src/index.js
```diff
@@ -40,8 +40,10 @@ document.addEventListener('DOMContentLoaded', () => {
     const { Router, createMemoryHistory } = require('react-router');
     window.router = createMemoryHistory();
     const routes = require('~/ui/routes');
+    const telemetry = require('~/telemetry');
 
     socket.start();
+    telemetry.init();
```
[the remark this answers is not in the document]
anri-asaturov: Telemetry needs send queue anyway, yeah waiting to render means you won't even see the login page if there's no internet.

suprko:
updated to new signup, pretty sure it's all working but it could use some more eyes @anri-asaturov @seavan @ebilg @kmist

src/telemetry/main.ts
```diff
@@ -0,0 +1,44 @@
...
+        return (...arg) => {
+            if (item(...arg)) send(item(...arg));
+        };
+    }
+    return () => {
+        send(item);
+    };
+}
+
+export function send(event) {
```
seavan: `event` is a global name in browsers, better use some other name

src/ui/signup/TermsDialog.jsx
```diff
@@ -0,0 +1,47 @@
+const React = require('react');
```
seavan: this looks like a duplicate of TermsOfUse.tsx

src/ui/signup/TermsOfUse.tsx
```diff
@@ -50,70 +60,75 @@ export default class TermsOfUse extends React.Component<SignupStep> {
     @action.bound
     selectTerm(number) {
         this.selectedTerm = number;
+        telemetry.signup.readMore(
+            this.termsTextContent[this.selectedTerm].titleLeft
+        );
     }
 
+    termsTextContent = [
```
seavan: this looks like reformatting but I am not sure. maybe checkout dev file and add telemetry so that the differences are more obvious?

suprko: Ah yeah before the array was just inside the `termsItems` getter and was directly mapped to the return. But I needed to have a named variable so that the titles could be passed to telemetry...

src/telemetry/components/login.ts
```diff
@@ -0,0 +1,61 @@
+import { telemetry } from 'peerio-icebear';
```
anri-asaturov: I think we use tsx for components

lostfictions:
@suprko I went ahead and improved the inference for the telemetry function -- please read it over and let me know if it works for you or if you want to discuss it.

suprko:
Ready for re-review, I think
@dchest @lostfictions @seavan @anri-asaturov

src/telemetry/components/signup.ts
```diff
@@ -0,0 +1,264 @@
+import { telemetry } from 'peerio-icebear';
+import { setup } from '~/telemetry/main';
+import { DurationEvent, TextInputEvent } from '~/telemetry/types';
+
+const { S, duration, errorMessage } = telemetry;
```
lostfictions: If you fix the `export default` issue I noted in icebear you should be able to just `import { S, duration, errorMessage } from 'peerio-icebear/telemetry'` instead of importing and then destructuring?

src/telemetry/index.ts
```diff
@@ -0,0 +1,14 @@
+import login from './components/login';
+import signup from './components/signup';
+import shared from './components/shared';
+
+function init() {
+    require('peerio-icebear').telemetry.init();
+}
+
+export default {
```
lostfictions: Again, please avoid `export default (anonymous new object)`

src/telemetry/components/signup.ts
suprko: `peerio-icebear/telemetry` gives me a "Cannot find module", I dunno if I'm doing something wrong

src/telemetry/index.ts
suprko: without a default export how do I `import telemetry from '~/telemetry'`?

suprko: `import  * as telemetry from '~/telemetry'` doesn't work, all of the functions are undefined

src/telemetry/components/login.ts
suprko: You mean `login.ts` should be TSX? Even though it doesn't use any JSX elements?

src/telemetry/index.ts
suprko: OK I must have missed an import. `import *` is working

src/telemetry/components/login.ts
anri-asaturov: Ah, it doesn't, I got distracted by the fact it's in 'components' folder.

anri-asaturov: Maybe it's better to move it out to logic folders.

suprko: what about calling the folder `events`?

anri-asaturov: Sure, let's just keep components in components folder and the rest out of it.

anri-asaturov:
conflict

ebilg:
can be merged

PR #382, opened 2018-07-23
feat: update emojione

anri-asaturov:


scripts/build-emoji.js
```diff
@@ -125,12 +125,12 @@ function convertToArrayAndCleanup(emojiJson) {
         item.ascii = item.ascii.slice(0, 2).join('  ');
 
         item.filename = `${relativePngFolder}${keys[k]}.png`;
-
-        item.characters = item.code_points.fully_qualified
+        item.characters = (item.code_points.fully_qualified || item.code_points.base)
```
anri-asaturov: Had to do this because in new data fully_qualified can be null

lostfictions:
ie., to fix:
```diff
diff --git a/test/unit/schema.test.js b/test/unit/schema.test.js
index a6b7c831..c8e09201 100644
--- a/test/unit/schema.test.js
+++ b/test/unit/schema.test.js
@@ -1,7 +1,7 @@
 const {
     chatSchema,
     isWhitespaceOnly,
-    emptyState
+    emptyDoc
 } = require('../../app/build/helpers/chat/prosemirror/chat-schema');
 
 
@@ -33,7 +33,7 @@ const paragraphWithEmoji = chatSchema.node('doc', null,
 
 describe('chat schema', () => {
     [
-        { doc: emptyState, name: 'empty doc', shouldBeWhitespaceOnly: true },
+        { doc: emptyDoc, name: 'empty doc', shouldBeWhitespaceOnly: true },
         { doc: hardBreakOnly, name: 'hard break only', shouldBeWhitespaceOnly: true },
         { doc: paragraphWithText, name: 'paragraph with text', shouldBeWhitespaceOnly: false },
         { doc: paragraphWithWhitespace, name: 'paragraph with whitespace', shouldBeWhitespaceOnly: true },
```

anri-asaturov:
I might have merged it with failing tests? Don't have other explanation.

lostfictions:
@anri-asaturov This is a type error that was introduced at some point, it just never got noticed or blocked anything because we don't currently run anything to check for type errors. My guess might be that in this case it was a transitive dep that got upgraded because we don't have a lockfile...?

scripts/build-emoji.js
lostfictions: hm, sure enough... was this always the case and we've been missing some emoji?

anri-asaturov: if you run the script you'll see the names, it seems that some older 'letter or number in a circle' emoji has been replaced with modern looking ones

lostfictions:
> @anri-asaturov This is a real type error that was introduced at some point, it just never got noticed or blocked anything because we don't currently run anything to check for type errors. My guess might be that in this case it was a transitive dep that got upgraded because we don't have a lockfile...?

@anri-asaturov Looks like it was you who broke it actually :P https://github.com/PeerioTechnologies/peerio-desktop/blame/dev/package.json#L69-L76

You can't upgrade `@types` dependencies like regular dependencies -- they typically track the major/minor version number of the package they provide typings for. (Patch versions can differ to allow for corrections in the typings themselves; sometimes minor version might differ under unusual circumstances but it's generally avoided.)

scripts/build-emoji.js
anri-asaturov: Does my replacement look right? `|| item.code_points.base`
I didn't dig into what it actually does, it just seemed to be the safe choice.

lostfictions: sure, presumably if there's no fully-qualified code points defined the base code points are what comprise it

ebilg:
receiving an emoji from the mobile app (👨🏾‍🌾) causes a blue screen logs attached
[-1532961116274.log](https://github.com/PeerioTechnologies/peerio-desktop/files/2241577/-1532961116274.log)

anri-asaturov:
@ebilg try again plz, it was fixed on dev, i just merged it

ebilg:
can be merged

PR #392, opened 2018-07-25
chore: upgrade prosemirror deps

lostfictions:
 
https://app.clubhouse.io/peerio/story/2208/update-prosemirror-related-dependencies

#### Testing instructions  

All package upgrades were minor so no breaking changes are expected -- except that the upgrade did create a minor breaking change in the schema. Should be backwards-compatible regardless, but viewing/sending rich text messages should be re-tested a bit just in case.

----
### Repository owner


ebilg:
I think this branch needs dev merged into it as well

anri-asaturov:
@ebilg done

PR #393, opened 2018-07-26
WIP: Files status window

suprko:
https://app.clubhouse.io/peerio/story/11543/desktop-file-status-ui
icebear: https://github.com/PeerioTechnologies/peerio-icebear/pull/264
 
#### Testing instructions  
markup only right now

----
### Repository owner

### DIFF
--- a/src/ui/files/components/FileStatusWindow.jsx
+++ b/src/ui/files/components/FileStatusWindow.jsx
@@ -1,0 +1,18 @@
+const React = require('react');
+const { observer } = require('mobx-react');
+
+@observer
+class FileStatusWindow extends React.Component {
+    render() {
+        return (
+            <div className="file-status-window">
+                <div className="title-bar">
+                    "File status (43 files)"
+                </div>
+                test
+            </div>
+        );
+    }
+}
+
+module.exports = FileStatusWindow;
--- a/src/ui/files/components/FileStatusWindow.jsx
+++ b/src/ui/files/components/FileStatusWindow.jsx
@@ -86,7 +86,7 @@ class FileStatusWindow extends React.Component {
         return (
             <div className={css('file-status-window', { collapsed: this.collapsed })}>
                 <div className="title-bar">
-                    <T k="title_fileStatus">{{ number: 43 }}</T>
+                    <T k="title_fileStatus">{{ number: FILES.length + FILES.length }}</T>
                     <div className="buttons-container">
                         <Button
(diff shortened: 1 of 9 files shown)


src/ui/files/components/FileStatusWindow.jsx
```diff
@@ -0,0 +1,157 @@
...
+    { name: 'file.jpg', iconType: 'img' }
+];
+
+@observer
+class FileStatusWindow extends React.Component {
+    @observable allCompleted = true;
+
+    @observable bodyRef;
+
+    setBodyRef = (ref) => {
```
lostfictions: You can just use `React.createRef()` (https://reactjs.org/docs/refs-and-the-dom.html#creating-refs) instead of both a field and a method definition -- it's more concise and it's the preferred way.

Refs should also probably be reference observables (https://mobx.js.org/refguide/modifiers.html) instead of full ones as well (tho they often don't need to be observables at all, eg. if you're just invoking a method on one -- looks like it does need to be observable here tho)

src/ui/files/components/FileStatusWindow.jsx
```diff
@@ -0,0 +1,157 @@
+const React = require('react');
```
lostfictions: Want to start adding `// @ts-check` in all new files you introduce?

src/ui/files/components/FileStatusWindow.jsx
```diff
@@ -0,0 +1,187 @@
...
+        return DOWNLOAD_FILES.map(f => this.fileItem(f, 'download'));
+    }
+    @computed
+    get uploadQueue() {
+        return UPLOAD_FILES.map(f => this.fileItem(f, 'upload'));
+    }
+
+    fileItem(file, actionType) {
+        const cancelFunction = `onCancel${actionType[0].toUpperCase() +
+            actionType.slice(1)}`;
```
seavan: ok, this was pretty funny to decipher :D @suprko 
but why just not have:

```javascript
    get downloadQueue() {
        return DOWNLOAD_FILES.map(f => this.fileItem(f, this.onCancelDownload));
    }
```

and same for upload?

suprko: lol. getting too fancy

ebilg:
@seavan is this ok to test now?

seavan:
@ebilg not yet it's just the markup, needs SDK from me.

PR #395, opened 2018-07-30
feat: beacons

suprko:
 https://app.clubhouse.io/peerio/story/13099/desktop-implement-area-beacon-component
https://app.clubhouse.io/peerio/story/10986/desktop-implement-spot-beacon-component-beacon-control
https://app.clubhouse.io/peerio/story/13245/desktop-implement-5-onboarding-beacons

#### Testing instructions  
The main beacon flow is intended to only show for new users immediately after signup. Also, once beacons are dismissed, their read status is saved in User's profile.

To make this easier to test, you can go to `ui-store.js` and setting `firstLogin = true`, and go to `beacon-store.ts` and comment out everything inside the `markAsRead` function. This will show you all beacons even when you're not on `firstLogin`, and it will avoid saving beacons to User profile so you can see them over and over again.

Otherwise the beacons (should) work as intended, which means most of them won't show unless you're on a completely new account on your first login after signup, and they won't reappear if you've dismissed them (or if you've taken certain other actions, e.g. navigating outside of where the beacon directs you results in all AppNav beacons being marked as read).

### Beacon sequences
Consult the Figma here for the intended flow: https://www.figma.com/file/gKlRNgEbAJGt6qNx4zrttR3B/%2311680-Desktop%3A-Initial-onboarding-Beacon-UI-placement I'll do my best to write out the flow below.

1. SpotBeacon on contacts icon in AppNav
  * After successful signup, user is directed to new Welcome screen. After 5 seconds of inactivity, beacon shows
2. AreaBeacon on user/email search in Contacts view
  * Shows if user has no contacts (whether new user or not)
3. SpotBeacon on files icon in AppNav 
  * If new user, shows after 5 seconds of user inactivity in Contacts view when viewing beacon # 2
4. AreaBeacon on "Upload" button in Files view
  * Shows if user has no files (whether new user or not)
5. SpotBeacon on chats icon in AppNav
  * If new user, shows after 5 seconds of user inactivity in Files view when viewing beacon # 4
6. AreaBeacon on plus icon on "Direct Messages" button in chat list
  * Shows if user has no chats or invites (whether new user or not)

Additional rules:
* A beacon that has been dismissed (click on the rectangle part of the beacon) will be marked as "read" and should not show again
* "New user" beacons apply only on the user's first use of the app after signing up. The `firstLogin` flag is set after a completed signup and will be reset if the user resets the app.
* If the user dismisses any of the AppNav beacons, or if the user navigates to the AppNav beacon in an order different than the one suggested by the beacons (e.g. clicking directly on chats when files is highlighted), then the AppNav beacons are all marked as read and don't show anymore.
----
### Repository owner


src/ui/shared-components/Beacon.jsx
```diff
@@ -0,0 +1,202 @@
...
+            className="beacon-container"
+            key="container"
+            ref={this.setContentRef}
+        >
+            {this.props.children}
+        </div>
+    );
+
+    // Beacon content needs to be `computed` because styles (positioning & sizing) can change based on observables
+    @computed
```
seavan: because beaconContent only gets called from `render` which only gets called when observables update, you can actually skip `@computed` here.

`@computed` is used to cache complex calculations, which may or may not include observables.

if you call `@computed` property twice with the same observables, you will get the same result the second time without invoking calculations.

if you only call this property when observables are updated (on render), @computed doesn't affect performance.

suprko:
Updated with beacon flow behaviour. @seavan @lostfictions thoughts on how I've set up the beacon flow, so we can have flow between components?

https://github.com/PeerioTechnologies/peerio-desktop/blob/07dcece8b110f805ab10bb112841a3331e9202a8/src/stores/ui-store.js#L38

In sum, each <Beacon> has an `activeId` which is a string. uiStore contains `beaconFlows` object, in which each item is an array of `activeId`s corresponding to a certain flow of beacons. When any beacon is clicked it will increment the `beaconNumber` observable which causes `currentBeacon` to change to the next `activeId`.

Inside <Beacon>, `active` computed turns true when this Beacon's `activeId` is equal to the `currentBeacon`, and `active=true` triggers the whole render sequence.

src/stores/ui-store.js
```diff
@@ -35,6 +35,22 @@ class UIStore {
...
+    @observable currentBeaconFlow = 'messageInput';
+    @observable beaconNumber = 0;
+
+    // Current beacon returns the appropriate entry from `beaconFlows`
+    @computed
+    get currentBeacon() {
+        if (this.beaconNumber < 0 || !this.currentBeaconFlow) return null;
+        return this.beaconFlows[this.currentBeaconFlow][this.beaconNumber];
+    }
+
```
anri-asaturov: I would do a separate class with clear api to safely control beacons,
here you need to change and reset variables externally, leaking logic.

src/ui/shared-components/Beacon.jsx
```diff
@@ -0,0 +1,255 @@
+// @ts-check
```
anri-asaturov: it's a new component, you can make it in tsx

src/ui/shared-components/Beacon.jsx
```diff
@@ -0,0 +1,255 @@
...
+                uiStore.beaconNumber + 2 >
+                uiStore.beaconFlows[uiStore.currentBeaconFlow].length
+            ) {
+                // // Reset beacon flow
+                // uiStore.beaconNumber = -1;
+                // uiStore.currentBeaconFlow = '';
+                uiStore.beaconNumber = 0;
+            } else {
+                uiStore.beaconNumber += 1;
+            }
```
anri-asaturov: yes, this is what i meant by the first comment, you should not do it here (change uistore variables), make a function you can call to do this for you

src/ui/shared-components/Beacon.jsx
```diff
@@ -0,0 +1,255 @@
...
+    // Render the child content, wrapped in .beacon-container div so we can make the above positioning calculations
+    childContent = (
+        <div
+            className="beacon-container"
+            key="container"
+            ref={this.setContentRef}
+        >
+            {this.props.children}
+        </div>
+    );
```
anri-asaturov: this is weird, why a class property?
It will initialize once and never change, meaning if props.children change - it won't, unless i miss something

anri-asaturov: getter is a good alternative

src/ui/shared-components/Beacon.jsx
```diff
@@ -0,0 +1,255 @@
...
+                </div>
+            </div>
+        );
+    }
+
+    render() {
+        if (!this.active && !this.rendered) return this.childContent;
+
+        const beaconContent = this.beaconContent();
+        return [this.childContent, beaconContent];
```
anri-asaturov: tricky.. was there no way to do a fixed position modal? Or is this easier?

anri-asaturov:
@suprko plz fix conflict before this branch rots

src/ui/shared-components/Beacon.jsx
suprko: I was debating between doing it like this or doing it the same way as Tooltip. Tooltip kinda cheats, you put it after the element that you want the Tooltip to appear on, and you hope (or make sure) that the parent is the same size as the element itself. It works though in a couple places there's some weird div nesting.

The thing is, both size and positioning for Beacon is much more precise than Tooltip, since the bubble visually wraps the element that it's highlight. So I felt like it was best to make the Beacon actually wrap the element that it highlights in the document. I don't know if this causes more problems than it solves though.

lostfictions: you can use a fragment instead of array and drop the keys, btw

src/ui/shared-components/Beacon.tsx
```diff
@@ -0,0 +1,271 @@
...
+            className="beacon-container"
+            key="container"
+            ref={this.setContentRef}
+        >
+            {this.props.children}
+        </div>
+    );
+
+    // Beacon content needs to be `computed` because styles (positioning & sizing) can change based on observables
+    beaconContent() {
```
lostfictions: At first glance I'd say, again, to split this up into a component for clarity rather than having "fat" getters... (also this comment... says it needs to be computed.... but then it's not marked as computed...?)

src/stores/ui-store.js
flohdot: @seavan had mentioned at some point doing something a bit like the top drawer is done on mobile. i'm not sure if that's still a useful thing to look at.

lostfictions:
Also, have you researched how eg. https://introjs.com/ and https://github.com/gilbarbara/react-joyride work? Any thoughts on those? I haven't used either of them and I have no idea whether they're well-architected, but they both do things like outlining/highlighting elements of arbitrary sizes -- without needing to modify the DOM tree. I think both use CSS selectors which is maybe not ideal, though.

src/stores/ui-store.js
lostfictions: Agreed that this is probably not what this should look like, imo

suprko:
@lostfictions Intro.js needs a subscription for commercial use so we probably rule that one out. Joyride looks cool, I'll have to dig in a bit and see how well it would play with a custom design. Targeting by selectors is not ideal but right now Beacons are targeted with an `activeId` prop, so that's ultimately kinda the same thing no?

As for DnD, is DragLayer the part I'm looking at? https://react-dnd.github.io/react-dnd/docs-drag-layer.html I can take see if this would be right for our purposes... it looks complicated though :confused:

lostfictions:
@suprko oops, sorry, wasn't clear -- i meant looking at the source for these, not looking into adopting them wholesale... i doubt intro.js would work for us regardless since it's imperative/dom-driven, but there might be hints about how it acquires element sizes, etc. without changing the enclosing tree.

joyride might be worth looking into as an option if that hasn't been discarded, i guess i just have generally low expectations of most js libraries that do a big thing unless they've got thousands of active users. but it may actually be good, or at very least there may be insights in the source.

src/ui/shared-components/Beacon.tsx
suprko: Is it any less cumbersome to make this a separate component if we have to pass it, by my count, 11 props?

anri-asaturov:
I have nothing against correct use of Portals

src/ui/shared-components/Beacon.tsx
lostfictions: Spread and rest operators exist :)

lostfictions: In this case I took a closer look and I think this is fine -- it comprises the main chunk of this component's render method anyway -- but I'm also not sure how much clarity you gain by having it as a separate method. It's totally up to you.

suprko: dw literally everything is changing due to the new Beacon type :grimacing:  actually i'm gonna close this PR until the restructure is done

suprko:
Reopened this WIP to get some opinions from ppl...

One of the biggest design challenges with the `SpotBeacon` component was the bubble -- how do we get this cutout bubble effect, where the beacon seems to have a cutout that makes a window to the content underneath? That's not really an easy thing with CSS. After a lot of back-and-forth with Paul and Skylar, Paul came up with a really neat idea where the bubble just appears to be punched out but actually has a solid background and we just match the content inside with the child content to make it look like the beacon has a window, when really the content has been copied.

This has made this component a lot easier to develop, but it makes implementation kinda finicky.

1. Child content is wrapped in a div with the `.beacon-container` class. This is necessary to accurately position the beacon. But there is the potential for this to interfere with the styling or positioning of the child content.

2. Child content is copied in the bubble of the Beacon itself. In some cases this will lead to us needing to copy the styles from the child content to the styles for the beacon bubble content. It's not difficult, but it is repetitive, which is obviously Bad Code...

IntroJS/JoyRide (linked above) both do a similar thing, which is that they show UI stuff on top of elements in the DOM without changing the DOM itself. They do this by targeting DOM elements with `id`s, which I don't actually think is so bad. This is totally fine for `AreaBeacon`, where it's just an arrow on top of the content. But for `SpotBeacon`, this brings us back to the initial issue, which is the difficulty of getting the cutout/window effect.

I suspect it is possible to do this effect with super precise use of `clip-path` and the same `gradient circle at` thing that's being used for the bubble's rounded corners. But then... we're back to the situation we started with, where I'd need a good chunk more dev time to actually build the beacon itself.

I'll talk to Paul and Skylar. In the meantime... what do devs think of this implementation? Too janky & annoying to work with?

anri-asaturov:
Could you show what you mean by cutout bubble?

seavan:
FYI @suprko @anri-asaturov we implement beacons by "faking" content in mobile. we don't have option of having cutout masks there.

suprko:
@anri-asaturov 
screenshot from 2018-09-13 16 04 48

anri-asaturov:
@suprko why can't it be a transparent div with image or css shapes in it?

suprko:
@anri-asaturov I'm not sure what you mean, is this different from using masks or clipping paths?

anri-asaturov:
You're just saying it's not easy with css, but there are many ways to do it, i'm trying to understand what's the issue. To me the idea with copying dom is way more complicated.

suprko:
Yeah I agree, that's why I'm trying to figure out what you're suggesting, because if this can be done without very precisely calculated clip paths or masks that would be way nicer. How would the div + image or CSS shapes work?

anri-asaturov:
ok
1. 
```
<beacon>
<left part with window style=transparent background>
<img src=png with transparency/>
</left part with window>
<right part with text>text</right part with text>
</beacon>
```

2.
```
<beacon>
<left part with window style=transparent background>
<div style=draw a circle with border, transparent background></div>
</left part with window>
<right part with text>text</right part with text>
</beacon>
```

anri-asaturov:
makes sense?

lostfictions:
@suprko one thing that'd be helpful for us to understand is maybe a deeper decomposition of the spot beacon and how its targeting/framing works exactly, ie. its sizing relative to the targeted object. if design has already laid that out in detail should we just check figma?

suprko:
@anri-asaturov `1` would require a different PNG for every combination of bubble/rectangle size and position... If the bubble was a static size this would be OK, we'd need `bubble positions * rectangle sizes` number of PNGs which would be manageable. What brings this down is that the bubble size scales according to content size, so there's infinite combinations... Though I can ask design if we actually need this. On this first set of beacons the spot beacon is apparently only used to highlight icons.

`2` is what I think we should do. It would require a clipping path to punch a hole in the rectangle exactly equal in size/position to the bubble sitting on top of it. I'm like 99% sure this would work but would take a lot of time time.

On that note, I didn't realize that Skylar was off for the next couple days... Paul said that it is almost guaranteed that Skylar wants us to get a working set of beacons into this upcoming release, and then we'll prioritize fixing it after the fact. So I guess that answers my question :man_shrugging:

suprko:
@lostfictions here's the docs & Figma for them:
https://docs.google.com/document/d/150qj4hYj0OMqdwMr2V12xVIL4cpr9mQzYRyimzwq2DA/edit
https://www.figma.com/file/QXNMvuHLXu5g1eCr9AZI2ux8/%231380%3A-Desktop%3A-design-beacon-component?node-id=72%3A1326

So yeah it looks like I need to get more detailed numbers from design. Most of the examples show the bubble wrapping an icon, one shows the bubble wrapping an avatar. Not a lot of detail on the amount of padding etc. if that's what you mean.

suprko:
@seavan 
screenshot from 2018-09-13 17 31 21

It's kinda hacky tbh.

```background: radial-gradient(
    circle at 50% 50%,
    transparent 50%,
    purple 51%
  );```

seavan:
@suprko wow. radial gradient. ok :) thanks!

lostfictions:
hey, here's a sketch of how i was imagining it could be structured: https://codesandbox.io/s/l4r4mwy0km (open in chrome, doesn't include a UA reset and it looks like ff has trouble with svg + viewport units)

the beacon definition itself might look a bit tricky but hopefully it's straightforward if you just gloss over the decorator stuff -- it provides a nice way for consumers to set up beacons externalizing all of the state and configuration in a way that's hopefully manageable, and within a component you can target a specific element for the beacon, which should auto-scale to match the size of the element. the general provider/decorator/consumer-and-portal structure is what i was suggesting when i meant that react-dnd would maybe be a useful reference for this stuff.

for a complex shape like this svg+masking is definitely the way to go imo, this sketch is just a quick pass but it could be refined further.

if it'd be helpful i could also sketch up or get on a call to discuss ideas as to how flow and state could work.

as an aside, from where i'm standing this is quite a complex feature -- i'd really strongly recommend against shipping a whatever thing just because skylar and/or paul would like it in the next release, unless there's a reason it's do-or-die? (speaking from experience here! i know folks were under a lot of pressure to get the files ui done, but later it was a pretty huge effort for me to refactor all of that just so that i could move on to implementing drag & drop... it'd be nice to avoid a repeat of that?)

i think it's worth thinking this through and getting it right. happy to take responsibility if a date slips because i suggested slowing down :)

suprko:
So the SVG version @lostfictions made in the sandbox is obviously way heckin better... But we have word From Above (like the highest level of Above) to cram whatever functional version that exists now into this week's release. This is clearly extremely not ideal, but this is what Above has requested.

FWIW these kinds of deadlines are something that I've been commenting on for awhile in my weekly reports... Maybe something to bring up at all hands.

BlueMona:
can be merged

PR #396, opened 2018-07-31
fix: in UserPicker, don't hide contacts when search results appear

suprko:
https://app.clubhouse.io/peerio/story/11777/desktop-continue-showing-other-contact-results-when-an-exact-match-is-found
 
#### Testing instructions  
user search in both new room and new DM should not hide existing contacts when exact match search result is returned

----
### Repository owner


ebilg:
can be merged

PR #398, opened 2018-08-01
Fix ability to write TypeScript files in this repo

lostfictions:

It turns out [`babel-eslint`](https://github.com/babel/babel-eslint) doesn't actually parse TypeScript correctly. It [doesn't actually read the user's `.babelrc` yet](https://github.com/babel/babel-eslint/issues/573); instead it has a hardcoded list of syntactic features it supports. It _looked_ like it was working for TypeScript because it's got Flow support built-in, but it explodes when using TS-specific syntax (eg. `const x = dog as Animal`). These failures are parsing errors, so it's not possible to suppress them with `eslint-disable-line`, either.

I attempted to switch to [`typescript-eslint-parser`](https://github.com/eslint/typescript-eslint-parser) instead, but it's got issues with non-TS code. I hit [tons of false positive `no-undef` issues for class members](https://github.com/eslint/typescript-eslint-parser/issues/471), for example. This rule has been a bit troublesome for the parser, but it's normally not a big deal for TS users since they can just disable it; the TS compiler already errors out for undefined values.

I opted to fix this by introducing (transitionally) an additional `.eslintrc` file that extends the main one but changes the parser to `typescript-eslint-parser` and disables the (now-unneeded but problematic) rules. The main eslint config now only runs for JS files, and the additional eslint only runs for TS files. This unfortunately means we lose in-editor linting for TypeScript files, but it can still be invoked manually and will run on CI. I looked into the [`vscode-eslint`](https://github.com/Microsoft/vscode-eslint) extension and it doesn't look like there's a way to get it to lint a subset of files in the same directory, so this seems like the best we can do.

Also fixed:
- Babel wasn't actually configured to check for or write .tsx files
- Some spurious or broken linter rules have been fixed
- `tsc` actually runs without errors now
- A validation script didn't have a hashbang and was trying to run under `sh` instead of `bash` in Linux
- Typescript version was upgraded; new version offers better error messages going forward.
- A few files converted to Typescript as a test.

Some of these changes should also probably be applied to icebear and mobile as well.

#### Testing instructions  

n/a.

----
### Repository owner


anri-asaturov:
Does it need testing or I can merge?

lostfictions:
This could probably use a scrutinizing eye (by someone who's not Anri :stuck_out_tongue:) and maybe just a bit of clicking around to double-check that I didn't do something totally silly, if @ebilg has a moment for it...

ebilg:
looks good

PR #399, opened 2018-08-02
chore: ts conversion, pt. 1

lostfictions:

First batch of files converted to TypeScript; functionality should be the same.
 
#### Testing instructions  

----
### Repository owner


src/helpers/keychain.ts
```diff
@@ -25,30 +34,24 @@ async function saveSecret(username, passphrase) {
...
         return Promise.resolve(true);
     } catch (err) {
-        console.error('Error getting passphrase with keytar', err);
+        console.error('Error deleting passphrase with keytar', err);
         return Promise.resolve(false);
     }
 }
-
-module.exports = {
-    saveSecret: Promise.method(saveSecret),
```
lostfictions: wasn't sure about this -- are these still necessary? the methods are already `async`, and `async` is already transpiled to Bluebird...

src/helpers/app-control.ts
```diff
@@ -1,25 +1,23 @@
...
-async function signout() {
+export async function signout() {
     const autologin = require('~/helpers/autologin');
     const { User } = require('peerio-icebear');
     await autologin.disable();
     await User.current.signout();
     relaunch();
 }
-
-module.exports = { relaunch, signout };
```
anri-asaturov: Isn't it nicer to see what's actually exported all in one place? (not for this 1-page file ofc)

src/helpers/keychain.ts
```diff
@@ -25,30 +34,24 @@ async function saveSecret(username, passphrase) {
...
+export async function getSecret(username: string): Promise<string | false> {
+    if (!keytar) return Promise.resolve<false>(false);
     try {
         const ret = await keytar.getPassword(service, username);
-        return ret || Promise.resolve(false);
+        return ret || Promise.resolve<false>(false);
     } catch (err) {
         console.error('Error getting passphrase with keytar', err);
-        return Promise.resolve(false);
+        return Promise.resolve<false>(false);
```
anri-asaturov: Hmm, is `false` a type as far as TS is concerned?

src/helpers/keychain.ts
anri-asaturov: no, they're not, we can search Promise.method and remove all

lostfictions: I guess we can also replace `Promise.resolve` in any async functions with just the values, also

src/helpers/app-control.ts
lostfictions: I don't have a super opinion, but what I'm doing is the convention almost all users of ES Modules seem to use.

I think it takes a bit of getting used to, but it actually makes reading a file nicer since you can know at the outset that what you're about to read is part of the interface for a module (instead of starting to read a file, having to drop to the bottom to see the exports, jump back up and check and compare...)

src/helpers/keychain.ts
lostfictions: Yes, most primitive literals are valid types, eg. `'a-literal-string'`, `303`, `false`. It's really valuable in a lot of scenarios.

src/main-process/context-menu.ts
```diff
@@ -39,7 +39,7 @@ function buildContextMenu(mainWindow) {
         });
     }
 
-    mainWindow.webContents.on('context-menu', (e, props) => {
+    mainWindow.webContents.on('context-menu', (_e, props) => {
```
anri-asaturov: What does the underscore do? Makes TS ignore unused var?

anri-asaturov:
ah test failed

src/main-process/context-menu.ts
lostfictions: Yeah, it's an assert against the compiler flag `noUnusedParameters`

anri-asaturov:
@lostfictions ok to merge?

lostfictions:
Should be ok

PR #403, opened 2018-08-08
chore(vscode): insert final newline, trim whitespace on save

dchest:
#### Testing instructions  

----
### Repository owner


anri-asaturov:
could you please check if prettier can do this? not everyone uses vscode

lostfictions:
We already have `.editorconfig` set up for this (which Prettier respects). Folks should make sure they have the EditorConfig extension enabled in Code instead:

screenshot from 2018-08-08 18-05-46

I'm not sure if having both the VSCode internal rules and EditorConfig will mess things up -- I assume it's okay for code files, but this PR might break this rule in `.editorconfig`:
```
[*.md]
trim_trailing_whitespace = false
```

@dchest Did you check that this isn't the case?

dchest:
```
[*.md]
trim_trailing_whitespace = false
```

Ah yeah, Markdown evil :) I'd prefer if was <br> instead of invisible double space.

PR #406, opened 2018-08-12
fix: file upload time formatting

suprko:
https://app.clubhouse.io/peerio/story/2146/client-dates-in-files-are-not-localized
icebear https://github.com/PeerioTechnologies/peerio-icebear/pull/276/
 
#### Testing instructions  
In Files view, file's "uploaded" date should display according to these rules:
* \< 1 day ago => relative time ("x minutes ago")
* 1 day ago => "Yesterday"
* \> 1 day ago => day of the week
* \> 7 days ago => full date in localized format ("July 24, 2018")

----
### Repository owner


src/ui/files/components/FileLine.jsx
```diff
@@ -98,21 +95,18 @@ class FileLine extends React.Component {
...
-                // this condition is always true and we need it only to subscribe to clock updates
-            } else if (uiStore.minuteClock.now) {
+            // If the file was uploaded more than an hour ago, just use the file's `uploadTimeFormatted` value.
+            if (Date.now() - file.uploadedAt > 60 * 60 * 1000) {
+                uploadedAt = file.uploadTimeFormatted;
+            }
+
+            // The below condition will always return true. We subscribe to minuteClock to force rerender of timestamp.
+            // i.e. if file was uploaded in the past hour, the "x minutes ago" will keep rerendering.
+            else if (uiStore.minuteClock.now) {
```
suprko: Unfortunately this is still necessary because `file.uploadTimeFormatted`, as a `computed`, will only change based on the value of the `observable` var `file.uploadedAt`. It won't update on its own on rerender (at least I couldn't think of a way to do it entirely in SDK). So we still need to subscribe to `uiStore.minuteClock` and recalculate the relative time on the desktop side. We do this only for files uploaded in the past hour.

seavan: you should add Clock.default.now observation into the computed in SDK and it will solve the problem.

icebear already has the Clock instance and everything is already happening there.

ebilg:
commented on ch

ebilg:
can be merged

anri-asaturov:
@ebilg was this not an issue?
https://github.com/PeerioTechnologies/peerio-icebear/pull/276#pullrequestreview-150088627

PR #410, opened 2018-08-22
feat: new Signup

suprko:
 https://app.clubhouse.io/peerio/story/12075/desktop-sign-up-ui-update

#### Testing instructions
New Signup flow!

* nice splash page with illustration should show if user has not previously created an account or logged in from this device
  * this will show for existing users until they log in once. may need some finessing?
* account creation process will suggest usernames to user based on their provided first and last name
  * username suggestions still kinda janky and will sometimes suggest invalid usernames (SDK fix incoming)
* account creation process now saves two user settings that it didn't before
  * subscribe to Peerio email newsletter
  * allow usage data collection
* throughout the process there's a "mock UI" on the right hand side that will update based on user input
* also check that things look good on a variety of window sizes. I've done a sweep already but worth another set of eyes.
  * big window
  * smallest window
  * wide and short
  * tall and narrow
* functionally nothing else is new, it's just split across more pages and visually nicer

----
### Repository owner


src/ui/signup/Signup.tsx
```diff
@@ -0,0 +1,143 @@
...
+                // todo: error message will not be localized, maybe don't use it at all
+                this.errorMessage = `${t('error_signupServerError')} ${
+                    errors.normalize(err).message
+                }`;
+            })
+            .then(async () => {
+                const { keyBackedUp } = this.profileStore;
+                keyBackedUp && (await User.current.setAccountKeyBackedUp());
+                // User.current.settings.dataCollection = this.profileStore.consentUsageData;
+                // User.current.saveSettings();
```
suprko: Uncomment these two lines to try saving dataCollection settings. Account creation itself will work but you'll get a snackbar and console errors that `User` settings did not save. This is exactly how `User` settings are saved in the Settings component itself so idk.

src/ui/signup/ProfileStore.jsx
```diff
@@ -0,0 +1,33 @@
...
+class ProfileStore extends OrderedFormStore {
+    @observable fieldsExpected = 4;
+
+    @observable username = ''; // also has observables usernameValid, usernameDirty
+    @observable email = ''; // also has emailValid, emailDirty
+    @observable firstName = ''; // etc
+    @observable lastName = ''; // etc
+    @observable passphrase = '';
+    @observable keyBackedUp = false;
+    @observable confirmedKeyBackup = false;
```
suprko: This is no longer actually used anywhere AFAIK but I got nervous about removing it for some reason. Safe to remove?

src/ui/login/Login.jsx
```diff
@@ -98,7 +99,12 @@ class Login extends Component {
                     .catch(() => {
                         this.loginStore.busy = false;
                     });
-            } else this.loginStore.busy = false;
+            } else {
+                this.loginStore.busy = false;
+                if (uiStore.newUserPageOpen) {
```
anri-asaturov: do you need this flag in uiStore tho? Can't you detect if we need to navigate to newUser right here?

src/ui/signup/CreateAccount.tsx
```diff
@@ -0,0 +1,454 @@
...
+                        ) : null}
+                    </div>
+                </div>
+                <div className="right-container">
+                    <div className="mock-app-ui">{content.right}</div>
+                </div>
+            </div>
+        );
+    }
+}
```
anri-asaturov: would be nice to break this down

src/ui/signup/ProfileStore.jsx
anri-asaturov: grep by name and see :-D

src/ui/login/Login.jsx
suprko: I tried some things but I couldn't find a way around manually setting a flag during a session...

For example, `if !previousUser => automatically redirect to NewUser page`
This makes it not possible to access the Login page. Because as soon as you open the Login page it redirects you back to the NewUser page again.

At one point I had NewUser as a modal within the Login page that could be shown/hidden based on a Login observable. But that introduced weird logic when trying to navigate directly to the Login page without seeing the NewUser page, and ultimately the same uiStore flag was needed anyway.

src/ui/signup/SignupStepTypes.ts
```diff
@@ -0,0 +1,9 @@
+export interface SignupStep {
+    store: any;
+    onComplete: () => void;
+}
+
+export interface StepContentObject {
+    left: React.ReactFragment;
+    right: React.ReactFragment;
+}
```
anri-asaturov: I personally am not in favor of typedefs disguised as components, maybe put interfaces in defs folder or smth

suprko: would we prefer a global defs folder, or a folder inside this component's folder?

anri-asaturov: i think a global folder would be easier to handle

PR #413, opened 2018-08-22
Typescript progress, #2

lostfictions:
#### Relevant info and issue/PR links 
 
Either merge this whole PR and close #411, or if you want smaller, more digestible chunks to review, #411 can be merged first, and then this can be rebased onto dev and reviewed separately.

I've tried to make additional commits consisting of only renames to facilitate git's rename tracking, and to make it easier to step through individual commits or groups of commits to follow renames. Please perform a regular merge instead of a squash merge to better preserve file history and the ability follow renames later on.

#### Testing instructions  

----
### Repository owner

- [ ] Was tested and can be merged.
- [ ] PR name follows conventional changelog format.

[PR guideline](https://github.com/PeerioTechnologies/peerio-icebear/blob/dev/docs/CONTRIBUTING.md)

PR #420, opened 2018-08-30
Typescript progress, #3

lostfictions:

Again, fast-forwardable from the previous PR (#413). I'm less confident about this one since it was spread out over several days and I've hit some typings bugs that I didn't want to fix/would rather suppress for now, but I _think_ it's maybe ok. (Some fixes to peer-ui will be incoming...)

#### Testing instructions  

----
### Repository owner


lostfictions:
@dchest would appreciate a quick eye on this too if you have time, though for sure icebear pr is more urgent

PR #429, opened 2018-09-18
style: add Fira Mono font, use for displaying account key

suprko:
https://app.clubhouse.io/peerio/story/13067/desktop-use-a-font-with-zeros-with-slashes-or-dots-for-displaying-account-key
 
#### Testing instructions  
In Signup flow, account key display uses Fira Mono now.
For consistency I also made Fira Mono the default monospace font (via `.monospace` class and `%monospace` selector), so it will also now be used in the account key display in Settings, user's "Peerio ID" in profile popup, there are probably some other places too.

----
### Repository owner


ebilg:
can be merged

PR #432, opened 2018-09-20
fix: delete received file from own Files view

suprko:
https://app.clubhouse.io/peerio/story/13945/desktop-delete-option-missing-from-three-dot-menu-for-files-received-in-dms
 
#### Testing instructions  
* User should be able to delete a received file (i.e. a file they do not own) from their own Files view. This will not remove the file from the DM in which it is received.
* The deletion confirmation message only asks to confirm that user wants to remove the file. "file will be unshared..." message should not show here

----
### Repository owner


src/ui/files/helpers/sharedFileAndFolderActions.ts
```diff
@@ -58,7 +58,7 @@ export function isFileShareable(file: File): boolean {
     return file.readyForDownload && file.canShare;
 }
 
-export function isFileDeleteable(file: File): boolean {
+export function isFileOwnedByCurrentUser(file: File): boolean {
```
suprko: since a file being deleteable is not necessarily the same as the file being owned by the current user, I renamed this

src/ui/files/components/FileActions.tsx
```diff
@@ -27,6 +27,14 @@ interface FileActionsProps {
...
     onMenuHide?: () => void;
 
+    /**
+     * By default, file is deleteable if `isFileOwnedByCurrentUser`.
+     * But in the Files view, user can delete a file they do not own.
+     * This only deletes their copy of the file and does not unshare it with others.
+     * `canDelete` can be used in this case to insist that deletion is possible even
+     * on a file the user does not own.
+     */
+    canDelete?: boolean;
```
suprko: this is a bit clunky, this kind of regresses to a previous era (individual `deleteable`, `shareable`, `moveable` props...), is there a better way...?

anri-asaturov:
eh, tbh this can be calculated in sdk, 
if file keg is in main store it can be deleted, otherwise it can be deleted if you're the owner.

anri-asaturov:
I've added File#deleteable computed property, just use it, it will work in 100% cases, it's in dev

src/ui/files/components/FileActions.tsx
```diff
@@ -142,7 +141,7 @@ export default class FileActions extends React.Component<FileActionsProps> {
                             onClick={this.props.onRename}
                         />
                     ) : null}
-                    {isFileDeleteable(file) ? (
+                    {this.props.file.deleteable ? (
```
suprko: @anri-asaturov `deleteable` returning true for non-owned files even inside DM...
screenshot from 2018-09-20 13 01 44

anri-asaturov: fixed

suprko:
OK, working now

anri-asaturov:
we got to fix the rest in similar manner, but not now, we can break something

ebilg:
looks good

PR #433, opened 2018-09-20
chore: bulk send telemetry events on signup completion

suprko:
https://app.clubhouse.io/peerio/story/13602/desktop-save-signup-telemetry-until-user-has-consented-then-bulk-send
https://app.clubhouse.io/peerio/story/7880/include-usage-statistics-in-error-reporting-setting 
icebear: https://github.com/PeerioTechnologies/peerio-icebear/pull/316

#### Testing instructions  
If the user is not signed up, telemetry events are not sent, should not be visible in Mixpanel. Instead, they are stored locally. Once user completes Signup and consents to sharing usage data, the events are sent in bulk.

----
### Repository owner


src/stores/telemetry-store.ts
```diff
@@ -0,0 +1,13 @@
+import { telemetry } from 'peerio-icebear';
+type EventObject = telemetry.EventObject;
```
suprko: @lostfictions post TS conversion I'm now able to import types from an icebear file. it's a little clunky though... what do you think?

lostfictions: yeah, this is why i was suggesting it would be ideal to have telemetry exported as a separate endpoint -- it's more of a namespace than an object.

@anri-asaturov was able to export model types as a separate endpoint under `peerio-icebear/dist/models` (which might become just `peerio-icebear/models` soon) -- maybe check that out.

lostfictions:
@suprko took more of a look at this today -- you're right, i think your solution is ultimately the simpler one. the one thing i think it could do is kind of "close the loop" of responsibility -- telemetry's `main` could either check whether there's stored events on a logged-in `send` call, or (probably better) just `autorun` `sendStored` on user becoming logged-in (and consent being true). That way consumers don't need to know that there's some invisible state that may be being accrued given some other opque-ish program state -- and don't need to know that it needs to be flushed with a separate, special method. What do you think?

src/telemetry/main.ts
```diff
@@ -51,10 +53,21 @@ function send(ev: TelemetryPayload): void {
         windowHeight: window.innerHeight
     };
 
-    const obj = {
+    const combinedProperties: telemetry.EventProperties = Object.assign(
```
lostfictions: Just use rest/spread

src/ui/login/Login.jsx
```diff
@@ -154,6 +154,14 @@ class Login extends Component {
         User.current = user;
         user.login()
             .then(() => {
+                when(
+                    () => User.current.settings.loaded,
+                    () => {
+                        if (User.current.settings.dataCollection) {
+                            telemetry.sendStored();
```
lostfictions: so just to confirm, i think `sendStored` doesn't need to be exported and both these calls (plus the additional proxying method declaration) can just be removed in favour of a single autorun in icebear/telemetry/main (or something similar)

lostfictions:
https://github.com/PeerioTechnologies/peerio-desktop/blob/8418e2ac40934b92bf1f9c2a4e2cbf86e6a83ae8/src/telemetry/components/shared.ts#L24

btw i think all `all`s related to `setup` should be removeable now

lostfictions:
looks good to me otherwise, pending some of the stuff in the corresponding icebear branch

PR #436, opened 2018-10-02
feat: system idle state detection

dchest:
When system is idle, enable mobile push notifications.
When it becames active, disable them.

It also detects locked/unlocked screen on Windows and Mac.

https://app.clubhouse.io/peerio/story/1938/recognize-desktop-inactivity-and-send-push-notifications-once-idle 

#### Testing instructions  

1. Set system sleep/screen lock time to be at least 15 minutes so that it won't interfere with testing.
2. Launch Peerio, open dev tools.
3. Wait for 5-8 minutes while doing nothing at the computer until `The system became idle. Enabled push notifications.` log messages appear. (Can make sure here if you receive mobile notifications, but it uses the same code as when system is going to sleep.) If they won't appear, it's a bug.
4. Wiggle mouse or press a key to came out of idle state. If you didn't interact with Peerio window (e.g. no mouse over it, or keyboard inputs), the active state will be detected in about 30 seconds from the point the system becomes active. Otherwise, if you do any input in Peerio (mouse, keyboard), it should be instant.
5. Check for `The system became active (not idle). Disabled push notifications.`

----
### Repository owner


src/helpers/power.ts
```diff
@@ -9,36 +10,82 @@ import { socket } from 'peerio-icebear';
...
+        await socket.send('/auth/push/enable');
+        console.log('Enabled push notifications.');
+    } catch (err) {
+        console.error(err);
+        console.log('Failed to enable push notifications.');
+    }
+}
+
+async function disablePushNotifications() {
     if (!socket.authenticated) return;
```
seavan: @dchest are you sure we don't need to wait for socket to be authenticated and do call /auth/push/disable? right now it looks like that if we have woken up and connection was disrupted, there's a good chance we won't be able to send disablePushNotificaiton because resumeHandler gets called before connection is established and ensured, no?

dchest: I've just refactored code that already was there, but you're right.

dchest: This needs when() and disposal of them... will fix tomorrow.

ebilg:
can be merged

PR #437, opened 2018-10-02
fix: make time formatter use new time zone if it changes

dchest:

Noticed that I still have Montreal time in message timestamps. This is because formatter is memorized for performance. This PR checks if time zone changed every hour and re-creates the formatter.
 
https://app.clubhouse.io/peerio/story/14212/time-zone-doesn-t-change-in-message-timestamps

#### Testing instructions  

----
### Repository owner


ebilg:
can be merged

PR #438, opened 2018-10-03
chore: Electron 3.0

dchest:

https://app.clubhouse.io/peerio/story/13900/electron-3-0

- [x] Update Electron
- [x] Rebuild node-keytar 
- [x] Fix extensions

#### Testing instructions  

* Check that every platform can load and store account key (i.e. automatic sign in works). There should be no errors in console about `keytar`.
* TBD

----
### Repository owner


dchest:
I think it's ready. Now sure if it should go into the next release. We have these two major changes that may cause something to break:

* Single package.json
* Electron 3.0

On one hand, it would be good to release them one-by-one: single package.json will go into the next release, and Electron 3.0 switch will go into the release after it so that if something goes wrong, we'll know where to look. However, since both switches require testing of everything, including updaters, maybe it would be a waste of time to test it twice, so maybe we should just release both changes in one release.

anri-asaturov:
merged this into https://github.com/PeerioTechnologies/peerio-desktop/pull/448
so Eren doesn't have to do regression testing twice.

anri-asaturov:
Reverting from #448 and reopening this until build is fixed

PR #446, opened 2018-10-08
WIP: E2E tests

alexandra-marin:
 https://app.clubhouse.io/peerio/story/6727/bootstrap-desktop-testing-framework
#### Testing instructions  

----
### Repository owner


seavan:
@alexandra-marin please fix conflicts 
I see several files fully commented out. maybe remove them? what's the reason to keep?

test/e2e/code/login.js
```diff
@@ -0,0 +1,38 @@
...
+    // console.log('My passphrase is', this.passphrase);
+
+    await this.app.client.waitForExist(withLabel('button_next'), 5000).click(withLabel('button_next'));
+    await this.app.client.waitForExist(withLabel('button_accept'), 5000).click(withLabel('button_accept'));
+    await this.app.client.waitForExist(withLabel('button_share'), 5000).click(withLabel('button_share'));
+});
+
+Then('I am taken to the home tab', async function() {
+    await this.app.client.waitForExist(withLabel(this.username), 5000);
+    await this.app.client.pause(5000)
```
lostfictions: This test seems incomplete?

test/e2e/code/login.js
```diff
@@ -0,0 +1,38 @@
...
+Then('I am presented with my passcode', async function() {
+    await this.app.client.waitForExist(withLabel('button_copy'), 5000).click(withLabel('button_copy'));
+    // .getText().then(innerText => {
+    //     this.passphrase = innerText;
+    // });
+    // console.log('My passphrase is', this.passphrase);
+
+    await this.app.client.waitForExist(withLabel('button_next'), 5000).click(withLabel('button_next'));
+    await this.app.client.waitForExist(withLabel('button_accept'), 5000).click(withLabel('button_accept'));
+    await this.app.client.waitForExist(withLabel('button_share'), 5000).click(withLabel('button_share'));
```
lostfictions: This test isn't asserting anything?

lostfictions:
@alexandra-marin, not sure if you wanted us to look at this immediately? For next time you make a PR, even if it's WIP for requesting feedback (which is totally fine and a good idea!) I think it'd maybe be helpful if you could leave a few notes in the PR description letting us know which parts are still WIP or what specifically you'd like feedback on? I made a few specific remarks above (and can add a few general notes below) but it's not clear whether this is working or what needs to happen for this to move out of WIP and into general review.

lostfictions:
I think this PR branch is based on #423 -- that was my suggestion, since it hadn't been merged yet and I thought we'd use Jest and maybe some Enzyme functionality for e2e tests. We could still use Jest's bundled `expect` assertion lib here, but otherwise this PR looks disjoint to me. So for clarity/readability in this PR, on top of deleting the old busted tests instead of commenting them out, I think this branch needs to be rebased to drop the commits already merged in from the other PR.

lostfictions:
re: using `data-` attributes -- it still feels a bit like an intrusive solution to me (every component in the UI library we want to test will need to be amended with a `testId` prop?)

Worse, it also feels like it still ends up being "implementation-specific" on a component-to-component basis (what does `testId` represent for the `Dropdown` component? For `RadioButtons`? Do we add multiple `testId` props to these components for outer/inner selectors?)

I wonder if we couldn't just use `className`, which is already implemented for most (all?) components in the UI library. As a bonus, offhand it seems easier to optimize away in production with a minifier (`css(IS_DEV_ENV && '_test-button-login', /* ...other selectors...*/)` or something similar). Thoughts?

test/e2e/code/login.js
```diff
@@ -0,0 +1,38 @@
...
+
+When('I choose the create account option', async function() {
+    await this.app.client.waitForExist(withLabel('signup'), 5000);
+    await this.app.client.click(withLabel('signup'));
+});
+
+When('I input my personal info', async function() {
+    this.username = new Date();
+    await this.app.client.waitForExist(withLabel('firstName'), 5000);
+    await this.app.client.setValue(withLabel('firstName'), 'Alice');
```
lostfictions: Having every step look like this is super noisy to me... is there any way `await this.app.client[A](withLabel(B), 5000)` could be reformulated more simply?

anri-asaturov:
I agree with @lostfictions, would be nice to either see a complete PR or comments on what to look at in the WIP.

anri-asaturov:
> re: using `data-` attributes -- it still feels a bit like an intrusive solution to me (every component in the UI library we want to test will need to be amended with a `testId` prop?)
> 
> Worse, it also feels like it still ends up being "implementation-specific" on a component-to-component basis (what does `testId` represent for the `Dropdown` component? For `RadioButtons`? Do we add multiple `testId` props to these components for outer/inner selectors?)
> 
> I wonder if we couldn't just use `className`, which is already implemented for most (all?) components in the UI library. As a bonus, offhand it seems easier to optimize away in production with a minifier (`css(IS_DEV_ENV && '_test-button-login', /* ...other selectors...*/)` or something similar). Thoughts?

What if className passed to some component will be applied to more than one child elements? Or I guess that's not usual. date attributes seem cleaner to me tbh.

There are some interesting comments in here
https://github.com/kentcdodds/react-testing-library/blob/master/README.md#getbytestidtext-textmatch-htmlelement
@lostfictions

anri-asaturov:
also this https://www.npmjs.com/package/babel-plugin-react-remove-properties

lostfictions:
@anri-asaturov yeah, makes sense to me -- i like kent's suggested guideline of preferring placeholder, label, alt text, text, etc. instead since it's closer to the way software is used by an end-user (and less invasive!). if we match localization keys in the tests maybe that's enough to get us most of the way there without instrumenting everything with test ids all of the time?

anri-asaturov:
Yeah, actually, locale keys are a great idea

anri-asaturov:
Fix conflicts please

lostfictions:
My fault on the conflicts, I suggested we base this branch off my branch #423 since it wasn't merged yet and I thought we would use Jest as a runner/assertions lib -- I think it's just confused with the squashmerge. @alexandra-marin you have some uncommitted/unpushed changes right? I can rebase this branch to drop the commits that are getting it confused now, but I don't want to make your local repo situation more confusing by pushing that -- let me know if you want me to do something.

lostfictions:
rebased

PR #448, opened 2018-10-10
Mobx5

anri-asaturov:


dchest:
Will need updater testing on Windows (2 staging releases, as usual).

PR #458, opened 2018-10-20
chore: update telemetry login events

suprko:
https://app.clubhouse.io/peerio/story/14752/desktop-add-telemetry-to-autologin-manual-login-with-and-without-2fa

#### Testing instructions  
Update to telemetry login events. ~~Kaoru is taking a look at the events themselves~~ [Tested & approved]. In the meantime devs could maybe take a look at some of the weird logic in here and see if it's OK?

There are some difficulties with the events that were requested. Mainly, 2FA and Login are independent components, like they basically don't have anything to do with each other (2FA is activated not from Login but by a call to server made in SDK). But design wants login events to know whether 2FA is enabled, and 2FA events to know whether autologin is enabled. So there are some weird `when`s being used to check these states and I dunno if it's overkill for like... 5 telemetry events. (Karim has done similar things on mobile).

----
### Repository owner


suprko:
Approved on the Mixpanel side

PR #459, opened 2018-10-22
fix: increase DMG window size

dchest:

https://app.clubhouse.io/peerio/story/15283/dmg-window-is-too-small-on-macos-mojave-with-scrollbars

#### Testing instructions  

Mac only:

1. `npm run internal-build-mac`
2. Open the latest `dist/something.dmg`.
3. Check that window size is acceptable.
4. Go to System Preferences > General, set Show scroll bars to Always.
5. Check that DMG window doesn't have scroll bars (can eject and re-mount it to be sure).

----
### Repository owner


ebilg:
can be merged

PR #475, opened 2018-11-03
refactor: AreaBeacons no longer needs to clone DOM

suprko:
https://app.clubhouse.io/peerio/story/14103/desktop-refactor-beacons-get-rid-of-dom-duplication

#### Testing instructions
Nothing should be any different. This change affects the way that area beacons (the ones with the circle) are constructed. It makes them less janky on the code side but should be visually identical to how they were before.

#### So what's the change?
One of the main reasons the current beacons are Bad™ is due to the fact that it's difficult to "punch out" something in pure CSS. The original solution to this was to copy the content that the beacon was wrapping, place it inside the bubble, and then match the bubble background to the original content's background. This obviously is super clunky and prone to error.

I later discovered a weird CSS trick where you can use `background: radial-gradient` to punch out a circle. The property lets you make an element's background colour a gradient of two (or more) colours which radiates in a circle from a defined center point. If you set the first colour to transparent, the second to a colour, and set the end of one colour and the start of the next just 1px apart, this effectively makes a solid transparent circle. Using the existing calculations for the bubble positioning, I've placed this circle at the exact size and position as the beacon's bubble. :sparkles:

I'm not sure if there are any disadvantages to this compared to recreating the shape in inline SVG? In any case, we can drop in an SVG beacon later; in the meantime this solves the issue of DOM duplication.

----
### Repository owner


src/ui/shared-components/Beacon.tsx
```diff
@@ -7,6 +7,7 @@ import { t } from 'peerio-translator';
 import beaconStore from '~/stores/beacon-store';
 
 const appRoot = document.getElementById('root');
+const BEACON_COLOR = '#5461cc';
```
anri-asaturov: I'm gonna pretend I didn't see this, cuz I don't have an easy solution to fix it.

suprko: yes it is extremely not ideal. unless we move to some kind of CSS-in-JS I don't see a way around this :dizzy_face:

anri-asaturov: reading `colors.scss` and parsing it? eh...

BlueMona:
Can be merged

src/ui/shared-components/Beacon.tsx
lostfictions: this is not whilelabellable as is then right?

suprko: It's not... also, we were never given a WL design for beacons so I don't actually know what colour they are on MC... :/

ebilg: our current plan is to do the same colour but that might change at some point

PR #478, opened 2018-11-06
improve localisation library

lostfictions:

ch story:
https://app.clubhouse.io/peerio/story/13663/improve-localisation-library

icebear branch:
https://github.com/PeerioTechnologies/peerio-icebear/pull/347

After merge, **all new imports of `t` should be from `peerio-icebear`, not `peerio-translator`**, so we should let folks know.

#### Testing instructions

No functionality should be changed.

New runtime warnings may be being generated for missing localization keys -- maybe click around a bit and see if you spot any after merge?

----
### Repository owner


lostfictions:
@anri-asaturov looks like it was your merge commit that broke this -- you might have to merge icebear dev into that feature branch, re-run the generate script there, and then re-run ci here

anri-asaturov:
fixed

lostfictions:
thanks!

lostfictions:
@anri-asaturov if it otherwise looks good, i'll fix the conflict and then this is ready for review

src/ui/chat/components/sidebar/MembersSection.tsx
```diff
@@ -72,7 +71,7 @@ export default class MembersSection extends React.Component<MembersSectionProps>
                     key={item.key}
                     value={item.value}
                     icon={item.icon}
-                    caption={t(item.caption)}
+                    caption={t(item.caption as any) as string}
```
anri-asaturov: hmm isn't `item.caption` a string?

src/ui/shared-components/Beacon.tsx
```diff
@@ -349,8 +349,9 @@ export default class Beacon extends React.Component<SpotBeaconProps | AreaBeacon
     }
 
     beaconContent() {
-        const title = this.props.title || t(`title_${this.props.name}_beacon`);
-        const description = this.props.description || t(`description_${this.props.name}_beacon`);
+        const title: string = this.props.title || t(`title_${this.props.name}_beacon` as any);
+        const description: string =
+            this.props.description || t(`description_${this.props.name}_beacon` as any);
```
anri-asaturov: i don't get the `any` cast, same as before, i guess

src/ui/chat/components/sidebar/MembersSection.tsx
lostfictions: so technically `keyof LocalizationString` is narrower than `string`... we could make `t` and `<T>` also accept any `string` as parameters, but that would mean, of course, that any string is a valid key. that is, we wouldn't be typechecking keys anymore. and we still wouldn't be able to infer whether the return type is also `string` or an array of mixed `React.children` elements.

unfortunately there are two casts needed -- one for the input value, and one for the output value, since otherwise i'll be an `any` that could leak elsewhere unless the field/parameter/binding/whatever it's being passed into has an explicit type (as i added for `description` below.)

this is obviously not ideal for the places where the key is assumed to be dynamic, but it reflects the fact that what the function accepts is not a `string` but a specific subset of string literals. icebear exports `LocalizationString`, so as much as possible things that will end up as localization keys should reflect that in their type. otherwise we're indeed introducing possible errors, and the casts are warning flags that we're doing something weird and understand the possibility of errors.

src/ui/shared-components/Beacon.tsx
lostfictions: fwiw i commented on the weirdness of this dynamic key assembly in lucas' recent pr -- i think interpolating the name when either a title or description are missing is a tricky bit of implicit behaviour. and again, the need for casts reflects that this might not necessarily be a safe thing to do.

src/ui/chat/components/sidebar/MembersSection.tsx
anri-asaturov: so we should basically type `item.caption` better

PR #481, opened 2018-11-07
WIP: beacons use Provider/inject instead of singleton store

suprko:
https://app.clubhouse.io/peerio/story/15152/desktop-refactor-beacons-move-from-ui-store-to-context-pattern

#### Testing instructions
Beacons should work the same as before, no change on the user side. Beacons that are currently on desktop:
* Welcome screen
  * spot beacon on Contacts icon
* Invite contact screen
  * area beacon on search bar
  * spot beacon on Files icon
* Files screen
  * area beacon on Upload button
  * spot beacon on Chats icon
* Chats screen
  * area beacon on plus icon of "Direct messages" button

----
### Repository owner


suprko:
I've made some more specific comments on individual files.

In general: this refactor essentially duplicates the existing logic that was in `beacon-store.ts`, but moves it to a new `BeaconContext` component, which uses mobx-react's `Provider`/`@inject` instead of the `export new BeaconStore` pattern. I did a bit of cleanup here and there but overall the logic is the same as before, for better or worse. I figure we can redo the logic as a separate task.

I'm sure there are some ways to tighten/simplify the logic here but I think it will always be really janky because we are dealing with such specific triggers. e.g. show the next beacon if the current beacon is 'search contacts', if the user has not touched their mouse or keyboard in 8 seconds, and if it's the user's first login to the app. So we're controlling the beacon with a mix of user input, global state, and component state... I don't know if there's a way that can be simpler or at least more centralized, but I am open to ideas!

src/ui/beacons/BeaconContext.tsx
```diff
@@ -0,0 +1,226 @@
...
+        clearIncrementQueue: this.clearIncrementQueue,
+        markAsRead: this.markAsRead
+    };
+
+    render() {
+        return (
+            <Provider
+                beaconInit={this.initializeBeacon}
+                beaconStore={this.beaconStore}
+                beaconsActive={this.activeBeacons}
```
suprko: `activeBeacons` => `beaconsActive` is kinda weird but hear me out...

I feel like `activeBeacons` makes more sense in this file (starting every var/function with `beacons...` would get confusing IMO). However, as far as I can tell, `Provider`/`@inject` occupies a single global namespace. It's not like React's context where you can specify `BeaconContext.Consumer`; with `inject`, the consumer just receives `@inject('propertyName')`. So we should probably have `inject`able properties start with the component they apply to, to avoid collisions, yeah?

src/ui/beacons/Beacon.tsx
```diff
@@ -0,0 +1,71 @@
...
+    'position',
+    'size',
+    'arrowPosition',
+    'arrowDistance',
+    'onBeaconClick',
+    'onContentClick',
+    'className'
+];
+
+@inject('beaconInit')
```
suprko: This is the part where I'm wondering whether I haven't actually solved the issue of the `beaconStore` pattern.

What happens is the `<Beacon>` component itself just wraps the child content, clones that content but with an added class `__beacon-target-id-beaconName`, and then passes relevant `props` up to the Provider with an "initialization" function. Provider now has an object with all of the beacons and their properties, which can then be passed to `BeaconItself` (which represents the actual beacon element as it appears in the DOM).

So like... is this still initialization? Is this the kind of thing we're trying to avoid? Or is it a bit better because it's still contained within the component-Provider chain rather than a global store? I guess I'm having a hard time seeing how this is better than a global store, but that could just be because I've implemented it in a `store`y way lol

src/ui/beacons/Beacon.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+        `React.Children.only` fails if children is >1 node (i.e. React.Fragment w/ many children)
+        This makes sense because we need to be able to measure the size of a *single* element
+        So this whole component will error if it wraps multiple children.
+    */
+    @computed
+    get clonedChildContent() {
+        const originalChild = React.Children.only(this.props.children);
+        return React.cloneElement(originalChild, {
+            className: css(originalChild.props.className, `__beacon-target-id-${this.props.name}`)
```
suprko: This is very ugly but makes accidental CSS targeting extremely unlikely. We could probably safely remove the double underscore.

src/ui/beacons/BeaconContext.tsx
```diff
@@ -0,0 +1,226 @@
...
+import routerStore from '~/stores/router-store';
+import {
+    addActivityListener,
+    removeActivityListener,
+    addActivityListenerWithoutMouseMovement
+} from '~/helpers/activity-listeners';
+import BeaconItself from './BeaconItself';
+
+@observer
+export default class BeaconWrapper extends React.Component<{}> {
```
suprko: I did my best to comment through all of these steps, but here's the entire beacon flow in one place to give you a broad idea if you're lost. Basically, these are the ways that beacons can work:

* `addBeacons` -> `increment`
  * this is the "normal" straightforward way for beacons to work
  * add all beacons to an array, `0`th element is the currently visible one, `shift` it to activate to the next beacon. simple enough.

* `addBeacons` -> `queueIncrement`
  * same as above but the component calls `increment` after a certain delay by using `queueIncrement`

* `queueBeacons`
  * similar to above, but instead of delaying the increment, you delay the `addBeacons` itself

(^ we could probably consolidate `queueBeacons` and `queueIncrement` into a single function come to think of it)

* `queueFirstBeacon`
  * the very first beacon in the app has a unique logic. this was directly requested by CEO. sorry.

suprko:
Actually making WIP because I realized I have to clean up some types. Comments still welcome though.

src/ui/beacons/BeaconContext.tsx
```diff
@@ -0,0 +1,226 @@
...
+                    <BeaconItself
+                        store={this.beaconStore}
+                        beaconsActive={this.activeBeacons}
+                        onIncrement={this.increment}
+                    />
+                </>
+            </Provider>
+        );
+    }
+}
```
lostfictions: This feels like quite a bit of code in general and state + state-munging functions in particular, for a single new file, especially a React component rather than an instantiable store... I haven't dug into the stateful parts yet, but at a glance it's not clear to me why the state is bound up in some kind of wrapper component rather than following a more typical store-provider-consumer pattern? Just noting this while reserving any deeper comment as I do a quick, shallow first-pass read of this...

src/ui/beacons/BeaconItself.tsx
```diff
@@ -47,25 +21,27 @@ interface RectanglePosition {
 }
 
 @observer
-export default class Beacon extends React.Component<SpotBeaconProps | AreaBeaconProps> {
+export default class BeaconItself extends React.Component<{
```
lostfictions: This is kind of funny in context and I understand why you arrived at this, but I think for later readers of this code it's going to be problematic and confusing. If you're struggling with naming now, with your more complete understanding of the problem at hand, it's very likely others (including your future self!) will be struggling with it even more later. In my opinions when you feel this way and don't have an imminent deliverable forcing you to ship whatever you've got, it's worth taking the time to try to see if you can figure out a better organizational principle.

src/ui/beacons/BeaconItself.tsx
```diff
@@ -47,25 +21,27 @@ interface RectanglePosition {
...
+    store: any;
```
lostfictions: Greenfield TypeScript code pointing at other TypeScript code should never be using `any` -- 99% of the time you can do better, and even if you're at the boundary of unknown data (I/O: filesystem, network, user input) you can lean on `unknown`. I'm not sure what this points to yet, but it probably shouldn't be annotated like this.

src/ui/beacons/BeaconItself.tsx
```diff
@@ -76,43 +52,50 @@ export default class Beacon extends React.Component<SpotBeaconProps | AreaBeacon
...
 
-    @observable dispose: IReactionDisposer;
-    @observable renderTimeout: NodeJS.Timer;
-    componentWillMount() {
-        // Update `contentRect` on window resize
-        window.addEventListener('resize', this.setContentRect);
+    // Properties of the current beacon, grabbed from `store`
+    @computed
+    get properties() {
+        return this.props.store[this.name];
```
lostfictions: This class seems to be pretty huge/complex, with almost everything being coupled to this `properties` getter (which incidentally doesn't strike me as doing important/costly enough work to merit being memoized like this, except to avoid a few characters of typing... First impression is that this class needs to be broken up quite a bit, but I haven't done a deep read of it yet.

src/ui/beacons/BeaconItself.tsx
```diff
@@ -76,43 +52,50 @@ export default class Beacon extends React.Component<SpotBeaconProps | AreaBeacon
...
-                if (active) {
-                    this.renderTimeout = setTimeout(() => {
+    fadeTimeout: NodeJS.Timer;
+    @observable reactionsToDispose: IReactionDisposer[];
+    componentDidMount() {
+        // (Re)calculate contentRect on window resize or beacon change
+        window.addEventListener('resize', this.setContentRect);
+        this.reactionsToDispose = [
+            reaction(
+                () => this.name,
```
lostfictions: This is what props are for :)

The `name` getter seems to be doing something decidedly un-namey; there's a lot of complex side effects going on here, where ideally this should be a component that just renders a reflection of some data. That doesn't necessarily preclude this from handling its stuff like counting the time it's been visible or (the canonical data-down, actions-up case) handling dom events, etc. But again, at first read there seems to be way too much happening here.

src/ui/beacons/BeaconItself.tsx
```diff
@@ -215,10 +198,10 @@ export default class Beacon extends React.Component<SpotBeaconProps | AreaBeacon
             There's also the "punchout" effect, created by placing a CSS punchout of the rectangle
```
lostfictions: This method is massive! Do others a mercy and extract it -- as a real function, with real parameters -- so it can be reasoned about in isolation. If you're really worried it's expensive, you can always make a memoization method that calls this other function and memoizes the result. But it's probably not as expensive as you think either.

src/ui/beacons/types.ts
```diff
@@ -0,0 +1,24 @@
+export interface BeaconBaseProps {
+    name: string;
+    title?: string; // if no title, will check t('title_${name}_beacon')
+    description?: string; // if no description, will use t('description_${name}_beacon')
```
lostfictions: These seem like somewhat tricky implicit behaviours for sort of marginal benefit; are you sure they're worth it?

This is also going make the typed translator functions complain -- rightly so, because they're not statically analyzable and thus more error-prone. If you want to be more explicit and maybe even win some type safety for the beacons, this could be a discriminated union that either takes a title and description, or else translation keys that are of type `keyof LocalizationStrings`.

src/ui/chat/ChatView.tsx
```diff
@@ -26,8 +25,9 @@ import ChatNameEditor from './components/ChatNameEditor';
 import ShareToChatProgress from './components/ShareToChatProgress';
 import PendingDM from './components/PendingDM';
 
+@inject('beaconActions')
 @observer
-export default class ChatView extends React.Component {
+export default class ChatView extends React.Component<{ beaconActions?: any }> {
```
lostfictions: See prev. comment about `any`.

I also thought we discussed this, but `inject` currently doesn't really do static typing well, so the new React context API is probably preferable. ([You can also use `static contextType` now](https://reactjs.org/blog/2018/10/23/react-v-16-6.html#static-contexttype) instead of a `<Consumer>` renderprop; I _think_ TypeScript should be smart enough to infer that it'll be available as a prop but I'm not 100% sure and you'll have to test.)

src/ui/chat/ChatView.tsx
```diff
@@ -50,13 +50,12 @@ export default class ChatView extends React.Component {
         ELEMENTS.chatView.checkActiveSpace();
 
         if (!chatStore.chats.length && !chatInviteStore.received.length) {
-            beaconStore.addBeacons('startChat');
+            this.props.beaconActions.addBeacons('startChat');
```
lostfictions: I might be misunderstanding this, but does this mean conditions for beacon activation aren't colocated with the beacon definitions, but instead are evaluated in individual components onmount? This seems a little confusing to me -- not ideal for readability or refactoring, and leaking details/concerns about beacons into components that shouldn't have to care. And what if there are beacons that might be activated from different components?

src/ui/chat/components/PlusIconBeaconed.jsx
```diff
@@ -1,6 +1,6 @@
 const React = require('react');
 const PlusIcon = require('~/ui/shared-components/PlusIcon');
-const Beacon = require('~/ui/shared-components/Beacon').default;
+const Beacon = require('~/ui/beacons/Beacon').default;
```
lostfictions: For small, trivial-ish files like this, especially with small numbers of dependencies/dependents, you might as well just do a quick conversion to Typescript if you can -- then you won't have to worry about whether imports need `.default` or not (and it'll help move the overall conversion along a bit!)

src/ui/contact/NewContact.jsx
```diff
@@ -73,13 +71,14 @@ class NewContact extends React.Component {
         this.notFound = false;
         this.query = newVal.toLocaleLowerCase().trim();
 
-        // Beacon management
-        if (beaconStore.activeBeacon === 'search') {
-            beaconStore.clearBeacons();
+        // Clear beacons after typing
+        if (this.props.beaconsActive[0] === 'search') {
+            this.props.beaconActions.clearBeacons();
```
lostfictions: Again, to me the action of typing in the textbox should probably just be dispatching some event to whatever is actually determining beacon visibility, not concerning itself with the logic of the beacons proper if it can avoid it.

src/ui/contact/NewContact.jsx
```diff
@@ -73,13 +71,14 @@ class NewContact extends React.Component {
...
         }
 
-        if (uiStore.firstLogin && !beaconStore.beaconsInQueue.length) {
-            beaconStore.queueBeacons('files', 8000);
+        // Queue the beacon on AppNav "files" button if it's the user's first session
+        if (uiStore.firstLogin) {
+            this.props.beaconActions.queueBeacons('files', 8000);
```
lostfictions: Ditto

lostfictions:
Incidentally, and this is a general process thing that I may have noted before, I worry it might be frustrating for everyone when a huge PR is lobbed over a fence one way and then a huge review, with a dozen or more long comments, is lobbed over the fence in return, repeated ad nauseam. I feel pretty strongly that it's worth having code that's works well and is as easy to understand as possible by everyone who might have to look at it, but I feel like that process shouldn't normally have to start with throwing rocks at the thing someone else just spent a lot of time working on.

We're lucky enough that our literal job descriptions that we get paid for include stuff like "ask questions", "answer questions", "work with other people to figure out problems", etc. :)  So please please don't feel like you're inconveniencing people if you're trying to think through a problem and aren't certain about something! You definitely don't have to wait until you've spent hours and hours on something and then add a postscript "I don't really know if this is better lol oh well", etc.

anri-asaturov:
I support this ^ you shouldn't think you're inconveniencing someone and all we expect here is an ideal code.

src/ui/beacons/BeaconContext.tsx
suprko: As mentioned above this isn't new code but mostly a copy & paste from the previous `beacon-store` singleton. My 1st goal was to replicate the existing functionality in a provider/context pattern, and later move on to a rebuild of the core beacon logic. But, we can do it all at once if that makes more sense.

suprko:
Maybe I misunderstood the nature of the task? What I thought I was doing:

1. get rid of DOM duplication (done in `refactor-beacons-css`)
2. avoid wrapping "beaconed" element in an extra `<div>`
3. move from the `beacon-store` singleton to provider/context pattern

So like, if I understood correctly, my task was just to get the beacons to exactly the same state they were in, under a provider/context paradigm, and with some of the DOM jank removed. A broader restructure of the beacon logic itself was not (I don't think???) in scope. Idk I could be wrong! If I was supposed to do a full rebuild of the beacon logic I definitely would have asked to pass it to (or at least pair with) someone else...

suprko:
Yeah that probably makes more sense

PR #485, opened 2018-11-09
refactor: Beacon remove div wrapper

suprko:
https://app.clubhouse.io/peerio/story/16965/desktop-refactor-beacons-to-remove-div-wrapper-on-beaconed-content

#### Testing instructions
No change on user side, beacons should work the same as before.

On code side:
Currently, beacons get their size/positioning from a div wrapper on the element they are highlighting. Obviously this causes problems with, for example, CSS targeting, because we're modifying the DOM. In this PR, the `<Beacon>` component returns a clone of the child content with an added "beacon-target" class, and then we query the DOM for that class and calculate based on that element. No wrapper needed.

----
### Repository owner


src/ui/shared-components/Beacon.tsx
```diff
@@ -76,19 +66,22 @@ export default class Beacon extends React.Component<SpotBeaconProps | AreaBeacon
...
 
     @observable dispose: IReactionDisposer;
     @observable renderTimeout: NodeJS.Timer;
-    componentWillMount() {
+    componentDidMount() {
+        // setContentRect on mount
+        this.setContentRect();
+
         // Update `contentRect` on window resize
         window.addEventListener('resize', this.setContentRect);
```
anri-asaturov: unsubscribe on unmount please

suprko: ? https://github.com/PeerioTechnologies/peerio-desktop/blob/04d15bec33c2e962352ffb3ee334921b226b36e7/src/ui/shared-components/Beacon.tsx#L97

anri-asaturov: hmm, couldn't find it in the diff

anri-asaturov:
tests need fixing

BlueMona:
Ok to merge

PR #486, opened 2018-11-09
chore: make all illustration images non draggable

suprko:
https://app.clubhouse.io/peerio/story/16879/desktop-illustrations-shouldn-t-be-draggable

#### Testing instructions
Illustrations in app should not be draggable. This is most noticeable in an empty folder, where dragging the background image causes the drag-and-drop upload UI highlight to show up. But in general it's something that can happen by accident and looks weird.

Affected illustrations:
* room invite
* chat zero state
* files zero state (including empty folder)
* contact list zero state and contact invite

----
### Repository owner


lostfictions:
ugh ci failure is the result of desktop riding `icebear#dev` -- forgot about that. @anri-asaturov i think all new desktop installs are going to fail until the desktop translator pr is merged :(  my fault for forgetting we ride the branch instead of versioning.

anri-asaturov:
Don't worry about that, it's a slow time and we just have released. Wednesday release will be from master anyway.

src/ui/chat/components/ChannelInvite.tsx
```diff
@@ -136,7 +136,7 @@ export default class ChannelInvite extends React.Component<ChannelInviteProps> {
         return (
             <div className={css('channel-invite', this.props.className)}>
                 <div className="invite-content">
-                    <img src="./static/img/illustrations/room-invite.svg" />
+                    <img src="./static/img/illustrations/room-invite.svg" draggable={false} />
```
anri-asaturov: does it have to be `{false}`? it's a html attribute

suprko: as in, why can't it be `'false'`? JSX expects a boolean

anri-asaturov: right, typescript sacrifice :)

anri-asaturov: nope, react unserstands `draggable="false"` fine

lostfictions: oh that's weird, i thought it usually didn't actually work with string attributes... are you sure it's not doing some weird coercion to falsy?

anri-asaturov: well ="true" works, so i guess no, it just knows how to handle it, OR, more likely, it just passes the string and DOM knows what to do with it. Although it does seem better type wise to use boolean values.

ebilg:
can be merged

PR #493, opened 2018-11-14
chore: use es module imports and exports almost everywhere

lostfictions:

No more `require('./local-file').default` confusion or errors! And should save significant time and effort for further TypeScript conversion work.

Everything seems to work fine -- the most tricky thing was a few instances where modules were used both as a namespace (eg. require the default, use members) in some places and as named imports in elsewhere.

I for the most part didn't attempt to convert or clean up any unusual uses of `require` (ie. anything not at the top level or with any additional kind of logic.) I cleaned up some imports of the translator urlMap from `config` (often for the better as `config` was being imported multiple times) and split some uses of `electron.remote` APIs into an import plus an assignment. A few imports were also rearranged for clarity as I went, too. (node builtins first, npm packages second, peerio support packages third, relative imports last, grouped by path.)

Source map support for the main process was added (only when running in dev env) to see accurate line mappings for transpiled files.

Some things that could be documented in Notion and/or posted in #news-desktop after merge:

- `module.exports` and `require` should generally not be used anymore, even in js code; even new conditional or unusual imports could maybe use `import()` syntax and/or the new `React.lazy` in some places to improve startup time/reduce jank.
- On desktop, `config` should be imported from `config.ts` at the src root rather than from icebear for consistency and typings compatibility/accuracy (even though they theoretically will do the same thing).
- It might be nice to discuss import order and enforce it via a lint (the [import/order](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md) rule) if everyone's on board.
- If you're in a file that now has an `export default X;` at the bottom, move it inline (ie. `export default class X { ...`) for consistency. (Or we could just fix them all via regex find-and-replace.)

#### Testing instructions

Everything will probably need to be tested, unfortunately, though I tried to work carefully and checked everything as I worked. On the bright side, this should make further changes less error-prone.

----
### Repository owner


anri-asaturov:
@lostfictions there are a few conflicts

lostfictions:
@anri-asaturov fixed

lostfictions:
fixed.... again.

@ebilg @BlueMona it'd be amazing if this could be tested sometime soon, it's going to accrue conflicts pretty much every time another PR is merged. as far as i can tell it's a safe change, but it's worth looking over just in case since it touches a lot of files.

ebilg:
it is next on our list!

lostfictions:
don't have time to fix conflicts again today :(  maybe thursday or friday

anri-asaturov:
I'll deal with it for you

PR #496, opened 2018-11-16
chore: PeerUI wrapper, default a11y styles on buttons/inputs

suprko:
https://app.clubhouse.io/peerio/story/12057/peerui-active-focus-feedback

#### Testing instructions
Basic keyboard navigation feedback. If you tab through elements in the app, buttons should highlight with a blue outline. If you move the mouse, the outline disappears.

This PR just changes the default styling, but doesn't change the app's own focusable elements. For example I'm not sure if the ChatList can even be tabbed through right now. That's for a more in-depth a11y sweep once designs are done.

For now the best place to test this is probably the signup process.

----
### Repository owner


src/ui/Root.tsx
```diff
@@ -113,7 +113,7 @@ export default class Root extends React.Component {
 
     render() {
         return (
-            <div>
+            <PeerUIWrapper>
```
anri-asaturov: What does it do? The name tells me nothing :(

suprko: Right now all it has is listeners that add/remove the `keyboard-nav` class based on user's keyboard or mouse input. I gave it a generic name because there might be other things in there later, e.g. theming

Full details here:
https://github.com/PeerioTechnologies/ui-library/blob/master/src/components/_Wrapper.tsx

anri-asaturov: Sounds sketchy :) @lostfictions you onboard?

suprko: already discussed :) https://github.com/PeerioTechnologies/ui-library/pull/22

anri-asaturov: Also mousemove handler on root... without debounce, looks like a bit too much. Is the problem this fixes worth it?

lostfictions: @suprko definitely think this needs a more specific name. and to stay single responsibility! we don't actually want a component that can do anything and have any functionality stuffed into it later imo.

lostfictions: @anri-asaturov please unfold the review comments in that PR, this was discussed. Lucas actually had throttled listeners set up, but since toggling to keyboard navigation or to mouse navigation is idempotent and early-outs, I actually recommended he remove them since it'll be more efficient.

The only potential risk is that it might thrash a bit if someone, like, somehow manages to hold the tab key down while simultaneously moving the mouse. I'm not sure that's a real concern.

lostfictions: @anri-asaturov you can also check https://github.com/PeerioTechnologies/ui-library/pull/21 for the alternative implementation Lucas proposed... i felt this approach was a lot better.

suprko: To be clear, the alternative implementation is no different in terms of the mousemove/keydown handlers. The only difference is how they're applied by the client code.

anri-asaturov: Yeah, ok looks good to me, read all the discussions.

ebilg:
😻 (can be merged)

PR #499, opened 2018-11-21
chore: use hook to remove link header instead of using a different fetcher

dchest:
#### Testing instructions

----
### Repository owner


dchest:
@anri-asaturov perhaps I should apply those commits to https://github.com/PeerioTechnologies/peerio-desktop/pull/500 instead of having a separate PR?

PR #500, opened 2018-11-22
feat: url preview

suprko:
https://app.clubhouse.io/peerio/story/17377/desktop-url-preview
icebear: https://github.com/PeerioTechnologies/peerio-icebear/pull/359

#### Testing instructions
URL previews working! The Figma mockups are linked in Clubhouse but there's a lot more in there than what we've implemented for this first MVP. Here are the possible scenarios:

1. direct link to an image file
  * show :framed_picture: icon 
  * show "Expand/collapse preview" text, which is clickable, as well as toggle button
  * preview just shows the image file, no website name etc.

2. link to website
  * show website's favicon (if available), otherwise :globe_with_meridians:  icon
  * show website's name (if available), otherwise hostname
  * if there is preview content, show expand/collapse toggle
  * preview content shows page title, description, and image

3. check that there are no CSP errors in console as in https://app.clubhouse.io/peerio/story/14391/url-unfurl-tries-to-load-css-blocked-by-content-security-policy (this PR has a different method of fixing that issue, the previous method was removed.)

----
### Repository owner


suprko:
@seavan I'm actually doing it right now, not as hard as I thought. the types are mismatched though

suprko:
OK, URL previews are working using PeerioTechnologies/peerio-icebear#359 :tada: 

So far, `siteName` and `title` are identical on all of the websites I've tried. I know OpenGraph has `og:site_name` but a lot of mainstream websites are not using that property. Worth bringing up with design?

dchest:
@suprko I've just made siteName be the domain name if it's identical to the title.

dchest:
screen shot 2018-11-26 at 15 36 46

Image sprite icon is 24x24, while favicon is 16x16 (as it should be), so image icon is being cut.

dchest:
BTW, we need to have some limits for siteName, title, and description, after which we cut them since this is external untrusted data. It can't go beyond a bit less than 64K chars (current limit for HTML), but I can image somebody putting a very long description just for the laughs.

suprko:
@dchest the only requirement I have from design is to limit the description to 280 characters. For title/sitename/etc. I just have height/width restrictions

dchest:
@suprko I'd file that under security requirement rather than design :-) I've added some reasonable truncation here: https://github.com/PeerioTechnologies/peerio-icebear/pull/359/commits/b5f05c5fc88003e8587173ee8aa9cc97f2d3e3c8 Will change limit for description.

dchest:
BTW, shouldn't image be clickable (pointing to the website?)

suprko:
Design wants it to eventually have a popup, the same as inline images.

suprko:
WIP removed, ready for review

src/ui/chat/components/UrlPreview.tsx
```diff
@@ -0,0 +1,209 @@
...
+}
+
+interface UrlPreviewImageProps {
+    type: 'image';
+    urlData: ExternalImage;
+    onImageLoaded?: () => void;
+}
+
+@observer
+export default class UrlPreview extends React.Component<
```
lostfictions: It sort of feels to me like there are two distinct use cases here, invoked in distinct contexts (ie. `externalImages` and `externalWebsites` are already two distinct props that are mapped separately in `Message.tsx`), but they're munged together in a single class here (which then has to do a bunch of branching and checks and ternaries to account for the separate cases... and then it ends up delegating to two distinct inner classes anyway).

Do you think it'd be a better approach to compose the common pieces within two distinct outer classes, no ternaries, rather than doing it like this? Eg. a `<UrlPreviewWebsite>` and `<UrlPreviewImage>` that can still share common markup (say, something like `<Foldable>` that accepts children).

src/ui/chat/components/UrlPreview.tsx
```diff
@@ -0,0 +1,209 @@
...
+    @action.bound
+    toggleExpand() {
+        this.expanded = !this.expanded;
+    }
+
+    @computed
+    get isToggleVisible() {
+        return (
+            this.props.type === 'image' ||
+            (this.props.type === 'html' &&
```
lostfictions: Just counting the number of places you have to branch on type...

1

src/ui/chat/components/UrlPreview.tsx
```diff
@@ -0,0 +1,209 @@
...
+                !!(this.props.urlData.title || this.props.urlData.description))
+        );
+    }
+
+    @computed
+    get headerText() {
+        if (this.props.type === 'html') {
```
lostfictions: 2

src/ui/chat/components/UrlPreview.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+    render() {
+        return (
+            <div className="url-preview">
+                <div className="url-info">
+                    <UrlPreviewHeader
+                        type={this.props.type}
+                        text={this.headerText}
+                        favicon={
+                            this.props.type === 'html' && !!this.props.urlData.favicon
```
lostfictions: 3

src/ui/chat/components/UrlPreview.tsx
```diff
@@ -0,0 +1,209 @@
...
+                        }
+                        isToggleVisible={this.isToggleVisible}
+                        onToggleExpand={this.toggleExpand}
+                        expanded={this.expanded}
+                    />
+                </div>
+
+                {this.expanded ? (
+                    <div className="preview-content">
+                        {this.props.type === 'html' ? (
```
lostfictions: 4

src/ui/chat/components/UrlPreview.tsx
```diff
@@ -0,0 +1,209 @@
...
+    text: string;
+    isToggleVisible: boolean;
+    onToggleExpand: () => void;
+    expanded: boolean;
+}> {
+    render() {
+        return (
+            <div className="url-preview-header">
+                <span className="favicon">
+                    {this.props.type === 'html' ? (
```
lostfictions: ...5.

imo when you see this happening over and over in a component you might have a case where the responsibility is inverted or the logic is misplaced, and there might ultimately be a simpler way to compose these things. (you might sometimes hear this referred to as "presentational" versus "container" components, often in the context of what has or doesn't have state, but i think it's a useful distinction for logic as well.)

(incidentally, i don't think branching or switching in a component is generally bad, though! i think it's mostly just a smell when you see logic scattered across a component over and over.)

ebilg:
can be merged

PR #504, opened 2018-11-28
feat: Chat and Files beacons

suprko:
https://app.clubhouse.io/peerio/story/13435/desktop-chat-beacons
https://app.clubhouse.io/peerio/story/13439/desktop-file-beacons
https://github.com/PeerioTechnologies/peerio-icebear/pull/364

#### Testing instructions
New beacons! Clubhouse has links to Figmas and GDoc spreadsheet. It's kinda spread out, here's the summary:

Chat beacons:
* highlight `+` share-to-chat icon if user has only 1 chat and has sent/received >5 messages
* highlith infopanel toggle once user has received at least 1 file in a DM
* highlight "pin" icon if user has more than 15 chats

Files beacons:
* when file is received via DM, and user later opens Files view, point to that file
  * if user has received multiple files via DM before seeing this beacon, the received file that's highest in the Files view will have the beacon
* highlight three-dots "actions" menu on the first ever file in Files view, no matter how it got there
* point to Add Folder button if user has 8 files and no folders

----
### Repository owner


src/ui/chat/ChatView.tsx
```diff
@@ -51,6 +58,10 @@ export default class ChatView extends React.Component {
         if (!chatStore.chats.length && !chatInviteStore.received.length) {
             beaconStore.addBeacons('startChat');
         }
+
+        if (!chatStore.myChats.favorites.length && chatStore.directMessages.length > 15) {
```
anri-asaturov: you don't need favorites in here, `directMessages` contains all DM chats

anri-asaturov: extract magic number to config plz

src/ui/chat/components/MessageList.tsx
```diff
@@ -76,13 +78,25 @@ export default class MessageList extends React.Component {
                 this.lastTopElement = null;
             }
         );
+
+        // If this is user's only chat, we show them the share-to-chat menu beacon
+        // after a certain number of messages have been sent or received.
+        if (chatStore.chats.length === 1) {
+            this.beaconReaction = when(
+                () => chatStore.activeChat && chatStore.activeChat.messages.length > 5,
```
anri-asaturov: Note that DM has 1 system message when created. Room has 2.
Extract magic number to config.

src/ui/files/Files.tsx
```diff
@@ -142,6 +162,35 @@ export default class Files extends React.Component<FilesProps> {
...
+    }
+
+    @computed
+    get recentReceivedFileIds() {
+        return chatStore.recentReceivedFiles.map(f => f.fileId);
+    }
+
+    @computed
+    get promptAddFolder() {
+        return fileStore.files.length >= 8 && fileStore.folderStore.folders.length === 0;
```
anri-asaturov: extract magic number to config

src/ui/files/helpers/fileBeaconContext.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export interface FileBeaconContextProps {
+    firstReceivedFileId: string;
+    firstListedFileId: string;
+}
+
+export const FileBeaconContext = React.createContext({
+    firstListedFileId: '',
+    firstReceivedFileId: ''
+} as FileBeaconContextProps);
```
anri-asaturov: Is this better than 'mobx-react' Provider/inject? @lostfictions
https://github.com/mobxjs/mobx-react#provider-and-inject

src/ui/shared-components/Beacon.tsx
```diff
@@ -64,39 +77,71 @@ export default class Beacon extends React.Component<SpotBeaconProps | AreaBeacon
...
+                this.scrollContainer = parentScrollRef;
+                this.scrollContainer.addEventListener('scroll', this.setContentRect);
+            }
+        }
+
+        this.contentRect = this.contentRef.getBoundingClientRect();
     }
 
-    @observable dispose: IReactionDisposer;
+    @observable reactionsToDispose: IReactionDisposer[];
```
anri-asaturov: does this need to be observable?

src/ui/shared-components/Beacon.tsx
```diff
@@ -64,39 +77,71 @@ export default class Beacon extends React.Component<SpotBeaconProps | AreaBeacon
...
     @observable renderTimeout: NodeJS.Timer;
     componentDidMount() {
         // setContentRect on mount
-        this.setContentRect();
+        setTimeout(() => this.setContentRect(), 0);
```
anri-asaturov: If you omit second argument - it means `0`

src/ui/files/components/FileLine.tsx
```diff
@@ -92,14 +95,44 @@ export default class FileLine extends React.Component<FileLineProps> {
...
+            onMenuHide={this.onMenuHide}
+            onActionInProgress={this.onActionInProgress}
+            onActionComplete={this.onActionComplete}
+            onDelete={this.deleteFile}
+            menuRef={this.fileActionsMenuRef}
+        />
+    );
+
+    @observable
+    fileActionsMenuRef: React.RefObject<Menu> = React.createRef();
```
lostfictions: it's both more concise and more typesafe if you omit the explicit annotation but use `React.createRef<Menu>()` instead.

src/ui/files/components/FileLine.tsx
```diff
@@ -92,14 +95,44 @@ export default class FileLine extends React.Component<FileLineProps> {
         this.contactProfileRef.current.openDialog();
     }
 
+    fileNameContent = (
+        <div
+            className="file-name"
+            onClick={this.props.clickToSelect ? this.toggleSelected : this.download}
```
lostfictions: this won't update if the props change! it's equivalent to declaring them in the constructor. it's sometimes ok to do this for totally static subtrees but not for anything depending on props (or observable values) -- even if it works right now for the current use-case it's extremely liable to cause bugs later when any part of this changes.

consider just introducing new components in the same file (they can be small, stateless functional components) rather than using getters or anything like that (they don't necessarily need to contain business logic like `this.props.clickToSelect ? this.toggleSelected : this.download`, so it's not as cumbersome and full of things to be passed down as it sounds). the benefit is that the main component definition stays small, but you still get the separation that, yes, these are individual subcomponents that can be reasoned about in isolation.

src/ui/files/components/FileLine.tsx
```diff
@@ -92,14 +95,44 @@ export default class FileLine extends React.Component<FileLineProps> {
...
+        >
+            {this.props.file.name}
+            {this.props.fileDetailsMini && <FileFolderDetailsRow file={this.props.file} />}
+        </div>
+    );
+
+    fileActionsContent = (
```
lostfictions: (ditto this one)

src/ui/files/helpers/fileBeaconContext.ts
lostfictions: yep. this has been discussed and the conclusion was that React's new context api is preferable; more typesafe, better-supported, etc.

(it seems like mobx-react might actually be slow-deprecating their own Provider/inject; their new prototype hooks version doesn't include them and recommends using React context)

lostfictions: incidentally i think just doing `React.createContext<FileBeaconContextProps>({ ... })` without the as-cast would work? using generics is generally preferable and less error-prone than as-casting when it works, even if the latter ultimately infers to the right result.

lostfictions:
didn't review the whole thing very closely -- but i guess i wasn't pinged for a full review anyway -- this is just the stuff that jumped out.

src/ui/chat/ChatView.tsx
lostfictions: btw, what's the logic on some of these things needing to be `when` and some being evaluable immediately? explanatory comments might be nice?

src/ui/chat/components/MessageList.tsx
suprko: The room/DM discrepancy is fine, the requirements gave a range ("show if 6-8 messages")

src/ui/files/components/FileLine.tsx
suprko: For now I've converted them to `computed get` but I'm wondering how to make these into separate components that won't need a lot of props?

Like you can see that `fileActionsContent` is just 
```
            <FileActions
                file={this.props.file}
                onMenuClick={this.onMenuClick}
                onMenuHide={this.onMenuHide}
                onActionInProgress={this.onActionInProgress}
                onActionComplete={this.onActionComplete}
                onDelete={this.deleteFile}
                menuRef={this.fileActionsMenuRef}
            />
```

and the only reason I assign it to a variable at all is just to avoid repeating that entire block twice (one inside a beacon, one without beacon). How can we make that a separate component without needing to repeat all of those props?

src/ui/files/components/FileLine.tsx
```diff
@@ -161,16 +211,33 @@ export default class FileLine extends React.Component<FileLineProps> {
...
+                            {context => {
+                                return (
+                                    <div className="file-actions">
+                                        {context.firstListedFileId ? (
+                                            <Beacon
+                                                name="fileOptions"
+                                                type="spot"
+                                                position="right"
+                                                description={t('description_moreFiles_beacon')}
+                                                onContentClick={this.onFileActionsBeaconClick}
```
lostfictions: okay, reading this over a bit more closely these are kind of gnarly... i think it's an ok solution for now but later implementation (or refactoring) of cases like this i wonder if we could extract all the beacon logic and markup into something that gets passed in as a prop, and has a default for when it's missing.

so like as a sketch in this case, instead of the component having the id passed in and then deciding whether to render the beacon over the actions -- and the same for the content, it gets passed in an id but then has all the logic and branching in here -- the parent decides for each of these cases whether the child should receive a function that wraps the content or the action (but each of these defaults to the identity function, ie. `i => i`). that would let us extract almost all of the beacon stuff into a separate component (that still receives all the info it needs) but that can still target a specific element in the child component. but it'd be a huge win in terms of not having the beacon implementations and logic leak into components like these and balloon their size and complexity. does that make sense?

@anri-asaturov what do you think?

lostfictions: lifting out the logic and markup seems like something that would be easy for the `receivedFile` beacon; the extra-gnarly thing for the `fileOptions` beacon is the click/ref forwarding stuff for sure. but maybe that could be expressed a little more simply.

src/ui/files/components/FileLine.tsx
lostfictions: right -- you're right, i didn't read this closely enough. i think the deeper problem is that there's branching/wrapping happening based on whether the beacon logic matches. i think a better solution could be to lift the logic out of this component and have a parent check and pass in either (a) if it matches, a beacon for the file actions or the filename, or (b) if not, an identity function.

but even if we don't want to do that now (you've already put a lot of work in here) i think you can simplify this a lot by, instead of extracting the inner components so you can repeat them on either side of a ternary or if-statement, extracting the beacon part and removing the branching altogether.

so
```jsx
@computed get fileNameContent() { ... }

[...]

<FileBeaconContext.Consumer>
                        {context => {
                            if (context.firstReceivedFileId === file.fileId) {
                                return (
                                    <Beacon
                                        name="receivedFile"
                                        type="area"
                                        arrowPosition="top"
                                        arrowDistance={20}
                                        offsetY={24}
                                        description={t('description_receivedFile_beacon')}
                                    >
                                        {this.fileNameContent}
                                    </Beacon>
                                );
                            }
                            return this.fileNameContent;
                        }}
                    </FileBeaconContext.Consumer>
```

would become something like

```jsx
static contextType = FileBeaconContext;

get maybeFileNameBeacon() {
  if (this.context.firstReceivedFileId === file.fileId) {
    return children => (
      <Beacon
        name="receivedFile"
        type="area"
        arrowPosition="top"
        arrowDistance={20}
        offsetY={24}
        description={t("description_receivedFile_beacon")}
      >
        {children}
      </Beacon>
    );
  }

  return children => children;
}


[...]

{this.maybeFileNameBeacon(
  <div
      className="file-name"
      onClick={this.props.clickToSelect ? this.toggleSelected : this.download}
   >
                {this.props.file.name}
                {this.props.fileDetailsMini && <FileFolderDetailsRow file={this.props.file} />}
  </div>
)}
```

indentation is a little messed up but hopefully that makes sense.

from there you can also easily see how it's trivial to extract that from a getter to its own function if desired, too.

lostfictions: also if you go a few steps further along the path of extracting this kind of thing to its own function... one thing you potentially end up with is a component where this `maybe` pattern (where it wraps children with _either_ a beacon, or nothing/identity) is reified, where the condition for deciding whether to render or not is just passed in as a prop. so if you see this happening a lot you could consider just rolling this into the existing `Beacon` or something.

src/ui/files/components/FileLine.tsx
anri-asaturov: I don't know tbh, I don't like everything that's related to beacons. It terrifies me so much i want to separate beacon implementation from all other component and implement them jquery-style :-D

lostfictions: clean separation from components was exactly what i was pushing for when it was being implemented, but the response i got was that we had to move forward with whatever because it had to ship. but that's not on lucas at all. selector-based overlays might ultimately be the cleanest way to do it, it's true.

anyway @suprko it's up to you whether you want to amend this with something like the alternate code i sketched in the other comment thread on this page -- we don't have to belabour this PR any more. but i think it's worth keeping that approach in mind next time you're faced with a situation like this where you either want to render `<B><A /></B>` or `<A />` (or to put it another way, `b(a)` or `a`).

src/ui/files/helpers/fileBeaconContext.ts
suprko: After playing with `inject` for a bit I think I like `Context` more. `inject` is too magical, like you just drop `@inject('firstThing', 'secondThing')` at the top of the class and it's not super clear where `firstThing` and `secondThing` are coming from. Importing a specific `Context` is clearer.

src/ui/files/components/FileLine.tsx
suprko: > we had to move forward with whatever because it had to ship

^ yeah this is exactly what happened.

This is indeed a very jQuery-ish kind of feature. I have that task for "refactoring" beacons (may be a complete rebuild?) but with current priorities it will probably keep getting bumped back. We can put a hold on more beacons (after these ones) until that gets done.

BlueMona:
Tested. All the beacons show up and are dismissed on action or clicking beacon. However, beacons do not disappear when the user navigates away. Also, the "infopanel toggle beacon" is supposed to show up when the first file is sent or received in a chat, but shows when received and not when sent.

BlueMona:
tested and can be merged

PR #514, opened 2018-12-06
e2e tests for creating chats

suprko:
https://app.clubhouse.io/peerio/story/17685/desktop-e2e-tests-create-a-room
https://app.clubhouse.io/peerio/story/17686/desktop-e2e-tests-create-a-dm

#### Testing instructions
Tests should be passing! Note that if you are using dev autologin (`autologin.json`) it will interfere with the tests, which require the app to start from zero user data.

----
### Repository owner


suprko:
`create-chat` tests are working :tada:

My main question is, how should we handle the fact that the app can do very different things based on a small piece of state?

e.g. just opening the app can produce three results:
  * existing user, autologin enabled => log in
  * existing user, autologin disabled => show login screen with "welcome back" message, username already filled in
  * no existing user => show NewUser screen with Sign In and Create Account buttons

I wasn't sure of the best way to build conditionals into these tests. Something like a `clickIfExists` function, with a timeout? Or, always make sure the app is booting fresh, with no previous data?

test/e2e/code/helpers/user.ts
```diff
@@ -0,0 +1,10 @@
+export const existingUsers = {
```
suprko: Wasn't sure if this was a good idea, since staging gets wiped sometimes... Is it better to just make new users on every test and treat those as "existing"?

test/e2e/spec/chat/create-dm.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Create new chat
+
+    Scenario: Create DM successfully
+        When I log in
+        And  I start a DM with a user
+        Then  I can send a message to the current chat
+        When  I navigate to Contacts
```
suprko: I changed this from `When I leave the chat` because "leave" is not a specific behaviour that a user can do (i.e. you can leave to any other place in the app -- Files, Settings, another chat...)

alexandra-marin:
LE: i see your tests passing on CI but when i run them locally they sometimes fail at room creation because start button in not clicked. code looks fine though ¯\_(ツ)_/¯

suprko:
Tests are passing locally and on CI.

The test user creation has been simplified a lot (thanks @alexandra-marin !) but it still requires an app restart. Directly creating a user via `ice` works and is a lot faster than going through the entire signup UI, but it still leaves bits of state everywhere and I couldn't figure out how to clear it without a proper app restart. If there is a way, let me know...

Otherwise, this works fine. I've added an explicit "create helper account" step as a `Background` in the feature (same as on mobile), which allows us to have a global-ish helper user that can be shared between steps. It looks kinda weird because the app will open and close multiple times, but it's still way faster than the previous version.

suprko:
@anri-asaturov this PR should fix the snackbar problem on the main `e2e` branch, so I think we can merge this into `e2e` -- and if that was the only problem with `e2e` then we can merge that to `dev`, what do you think?

src/ui/chat/components/MessageInputProseMirror.tsx
```diff
@@ -341,6 +341,7 @@ export default class MessageInputProseMirror extends React.Component<{
                         className="message-editor-container-prosemirror"
                         onBlur={this.onInputBlur}
                         ref={this.mountProseMirror}
+                        data-test-id="messageInputContainer"
```
anri-asaturov: why the inconsistent `testId/data-test-id`?

suprko: `testId` on React components, `data-test-id` on HTML elements

PR #516, opened 2018-12-11
chore: beacon logic fixes

suprko:
https://app.clubhouse.io/peerio/story/18025/desktop-don-t-show-chat-pin-beacon-to-old-users-if-they-already-have-pinned-chats
https://app.clubhouse.io/peerio/story/18029/desktop-infopanel-beacon-should-not-show-for-existing-users

#### Testing instructions
* "Infopanel" beacon logic has been very restricted. It will now only show if you have sent/received a file in the current chat *and* there is only 1 file (the one you just sent/received) in your fileStore. So, for example, it will *not* show if
  * you have uploaded files on your own and then send/receive a file in a chat
  * you have already received a file in `chat A` and then you send/receive a file in `chat B` without looking at `chat A` first
  * you are offline when you receive more than 1 file
* "Pin chat" beacon will not show if the user already has 1 or more pinned chats.

----
### Repository owner


src/ui/chat/ChatView.tsx
```diff
@@ -61,7 +61,10 @@ export default class ChatView extends React.Component {
             beaconStore.addBeacons('startChat');
         }
 
-        if (chatStore.directMessages.length > config.beacons.dmCountPinPrompt) {
+        if (
+            !chatStore.myChats.favorites.length &&
```
lostfictions: if this is supposed to mean "length is zero", please just use "length is zero" :)  no need to be clever to save a few characters!

src/ui/chat/ChatView.tsx
```diff
@@ -44,7 +44,7 @@ export default class ChatView extends React.Component {
                 }
             ),
             when(
-                () => chatStore.activeChat && chatStore.activeChat.recentFiles.length > 0,
+                () => chatStore.activeChat.recentFiles.length === 1 && fileStore.files.length === 1,
```
lostfictions: can `chatStore.activeChat` be null/undefined? it was being checked before but isn't anymore...

anri-asaturov: It can be, yes.

BlueMona:
Tested December 19th. Bugs on CH. 

https://app.clubhouse.io/peerio/story/18128/desktop-pin-beacon-shows-after-user-pins-unpins-and-signs-back-in

https://app.clubhouse.io/peerio/story/18123/desktop-infopanel-beacon-shows-even-when-user-receives-multiple-files-while-offline-or-has-other-files-in-file-store

BlueMona:
Tested and can be merged (pending checks?)

PR #518, opened 2018-12-19
feat: illustration on DM start with 0 contacts

suprko:
https://app.clubhouse.io/peerio/story/17897/desktop-illustration-for-new-dm

#### Testing instructions
Illustration has been added to new DM screen, only when user has 0 contacts.

I also made a simple <BackgroundIllustration> component for this type of image, which I think we're expecting more of. Changed the NewContact illustration to use this component too, so that should be checked to make sure that it still works fine.

I had to make a small adjustment to the UserPicker height (using `flex` instead of `vh` unit). I did a quick look to make sure that it wasn't broken but a second set of eyes would be good. UserPicker is used here:
* new DM
* new Room
* add member to Room

----
### Repository owner


BlueMona:
Tested and can be merged.

PR #527, opened 2019-01-03
feat: remember opened folder when switching to/from Files view

suprko:
https://app.clubhouse.io/peerio/story/14684/desktop-remember-folder-when-switching-back-to-files-tab

#### Testing instructions
Go to Files, open a folder, navigate away from Files and then back again, same folder should remain open.

----
### Repository owner


src/ui/files/components/FilePicker.tsx
```diff
@@ -26,13 +26,26 @@ export default class FilePicker extends React.Component<FilePickerProps> {
...
     pageSize = DEFAULT_RENDERED_ITEMS_COUNT;
 
+    setOpenedFolder: IReactionDisposer;
+    previousOpenedFolder: FileFolder;
+
     container: HTMLElement | null = null;
 
     componentWillMount() {
-        fileStore.folderStore.currentFolder = fileStore.folderStore.root;
+        this.setOpenedFolder = when(
```
anri-asaturov: AFAIR this component can be mounted and then become active/inactive multiple times, meaning that your `when` will only run once.

suprko: Yeah but we only need to save the `openedFolder` (the folder that was opened in Files view) once, right? That's not going to change if FilePicker is opened/closed multiple times.

anri-asaturov: It's better to assume that you(Picker) don't know what can and can't happen with `currentFolder`, it can easily be changed if not now, then in future, while the dialog is mounted.
All the picker cares about is not to break `currentFolder` when it opens and closes.

suprko: OK done

ebilg:
can be merged

PR #528, opened 2019-01-04
feat: drag and drop style update in chats & allow dropping onto contacts in contacts view

lostfictions:

https://app.clubhouse.io/peerio/story/15513/desktop-drag-and-drop-style-update

This PR removes the old `DropTarget` in favour of a unified UI in the style already used for drag and drop in the Files pane.

I've also refactored `ChatView` so that it's three components and `ContactList` so that it's two; all the new components should be testable with Enzyme/whatever (but new tests are not included in this PR).

#### Testing instructions

Most stuff relating to the contact list and uploading files (both drag and drop and via menu) should be tested if possible!

----
### Repository owner


src/ui/contact/ContactList.tsx
```diff
@@ -0,0 +1,136 @@
...
+import routerStore from '~/stores/router-store';
+import UploadDialog from '~/ui/shared-components/UploadDialog';
+import T from '~/ui/shared-components/T';
+
+import ContactListItem from './ContactListItem';
+
+@observer
+export default class ContactList extends React.Component {
+    readonly sortOptions = [
+        { value: 'firstName', label: t('title_firstName') },
```
lostfictions: belatedly realizing this needs to be a getter or it's not going to change when the locale does

ebilg:
commented on CH

BlueMona:
commented on CH

BlueMona:
tested for fix. Ok to merge.

